PR #3, opened 2025-05-29
Convert legacy publicKey* formats in createDidDocument to use publicKeyMultibase

venables:
In the core [DID v1.0 spec](https://www.w3.org/TR/did-1.0/#verification-material), the only supported formats for a public key in a `verificationMethod` are `publicKeyMultibase` and `publicKeyJwk`. 

> To increase the likelihood of interoperable implementations, this specification limits the number of formats for expressing verification material in a DID document.

This PR updates the `createDidDocument` method to convert publicKeyHex and publicKeyBase58 formatted publicKeys to use `publicKeyMultibase` in the document.  Multibase is chosen because it is the same fundamental representation as Hex/Base58 (string versus an object in Jwk)

### DIFF
--- a/.changeset/smart-buckets-end.md
+++ b/.changeset/smart-buckets-end.md
@@ -1,0 +1,9 @@
+---
+"agentcommercekit": minor
+"@agentcommercekit/keys": minor
+"@agentcommercekit/did": minor
+---
+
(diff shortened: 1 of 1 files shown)


domleboss97 (approved):
Looks good!

PR #12, opened 2025-06-25
Update A2A demo flow to include VC exchange

domleboss97:
Updates the A2A demo flow to include exchange of client and server VCs in handshake messages.


venables (approved):
Looks great!

PR #15, opened 2025-06-26
Add ES256 support 

venables:
- Add support for secp256r1 (P-256) elliptic curve
- Simplifies the DID `verificationMethod` types to be either `Multikey` (for Multibase keys) or `JsonWebKey2020` (for JWK keys).
- Adds schemas for our jwt payload and jwt headers


domleboss97 (approved):
Looks good!

PR #22, opened 2025-07-04
Update A2A demo to use official @a2a-js/sdk package

venables:
The demo was using `a2a-js` which is a 3rd party package. This updates the demo to use the official `@a2a-js/sdk` package.

### DIFF
--- a/packages/ack-id/src/a2a/sign-message.ts
+++ b/packages/ack-id/src/a2a/sign-message.ts
@@ -123,14 +123,12 @@ export async function createA2AHandshakeMessage(
 
   const message = createA2AHandshakeMessageFromJwt(role, jwt)
 
-  const a2aHandshakeMessage: A2AHandshakeMessage = {
+  return {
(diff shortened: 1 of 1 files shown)


packages/ack-id/src/a2a/verify.ts
```diff
@@ -62,10 +62,11 @@ export async function verifyA2ASignedMessage(
   { did, counterparty, resolver = getDidResolver() }: VerifyA2AHandshakeOptions
 ): Promise<JwtVerified> {
   // Ensure the message is a valid A2A signed message
-  const { metadata, ...parsedMessage } = v.parse(
-    messageWithSignatureSchema,
-    message
-  )
+  const {
+    metadata,
+    contextId: _contextId,
```
edspencer, line 67: doesn't appear to be used

edspencer (approved):
Couple of small things but otherwise LGTM

PR #23, opened 2025-07-09
Add methods to check if a public key is valid for a given curve

venables:
Adds `isValidPublicKey(pubkey: Uint8Array): boolean` method to each of the supported curves which checks if the key length is valid and if the point falls on the curve.


edspencer (approved):
LGTM

PR #24, opened 2025-07-17
Replace StatusList2021 (deprecated) with BitstringStatusList

edspencer:
Replaces StatusList2021 with [BitstringStatusList](https://www.w3.org/TR/vc-bitstring-status-list/). Logic remains effectively the same, just some minimal schema changes.

### DIFF
--- a/examples/verifier/README.md
+++ b/examples/verifier/README.md
@@ -7,8 +7,6 @@ This example showcases a **Credential Verifier** for [ACK-ID](https://www.agentc
 
 This verifier uses [Bitstring Status List](https://www.w3.org/TR/vc-bitstring-status-list/), to check if a credential is revoked.
 
-# Installation
-
 ## Getting Started
 
 ```sh
--- a/packages/vc/src/schemas/valibot.ts
+++ b/packages/vc/src/schemas/valibot.ts
@@ -36,7 +36,7 @@ export const jwtProofSchema = v.object({
   jwt: v.string()
 })
 
-export const bitstringStatusListCredentialSchema = v.object({
+export const bitstringStatusListClaimSchema = v.object({
   id: v.string(),
   type: v.literal("BitstringStatusList"),
   statusPurpose: v.string(),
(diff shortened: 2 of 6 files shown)


examples/verifier/README.md
```diff
@@ -5,7 +5,9 @@ This example showcases a **Credential Verifier** for [ACK-ID](https://www.agentc
 - `ControllerCredential`: ACK-ID credentials that prove DID ownership hierarchies.
 - `PaymentReceiptCredential`: ACK-Pay credentials that provide proof of payment that satisfies a given Payment Request.
 
-This verifier uses [StatusList2021](https://www.w3.org/community/reports/credentials/CG-FINAL-vc-status-list-2021-20230102/), to check if a credential is revoked.# Installation
+This verifier uses [Bitstring Status List](https://www.w3.org/TR/vc-bitstring-status-list/), to check if a credential is revoked.
+
+# Installation
```
venables, line 10: Was this added on purpose? I think we have an H1 already

venables (approved):
Looks good, I think just that README update.  

As we release this we should note in the release notes that this is a breaking change

packages/vc/src/schemas/valibot.ts
```diff
@@ -36,9 +36,9 @@ export const jwtProofSchema = v.object({
   jwt: v.string()
 })
 
-export const statusList2021ClaimSchema = v.object({
+export const bitstringStatusListCredentialSchema = v.object({
```
domleboss97, line 39: I think maintaining calling this `...ClaimSchema` makes more sense , since this is the schema just for the claims within the credential, not the whole credential itself. This happens in some other places as well.

domleboss97 (approved):

packages/vc/src/schemas/valibot.ts
edspencer: Ok makes sense, I've updated it to do so 👍

PR #29, opened 2025-08-12
Improve JWK types and methods, add did:jwks support

venables:
- Adds methods to allow more explicit identification of publicKey-only JWKs vs JWKs containing a private key.
- Adds a `did:jwks` resolver

### DIFF
--- a/packages/did/src/create-did-document.ts
+++ b/packages/did/src/create-did-document.ts
@@ -2,6 +2,7 @@ import { encodePublicKeyFromKeypair } from "@agentcommercekit/keys"
 import {
   base58ToBytes,
   bytesToMultibase,
+  getPublicKeyJwk,
   hexStringToBytes
 } from "@agentcommercekit/keys/encoding"
 import type { DidDocument } from "./did-document"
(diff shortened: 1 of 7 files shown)


packages/did/src/did-resolvers/get-did-resolver.ts
domleboss97, line 1: love it

packages/did/src/create-did-document.ts
```diff
@@ -49,7 +50,7 @@ export function createVerificationMethod({
   switch (encoding) {
     case "jwk":
       verificationMethod.type = "JsonWebKey2020"
-      verificationMethod.publicKeyJwk = value
+      verificationMethod.publicKeyJwk = getPublicKeyJwk(value)
```
domleboss97, line 53: is there any way to enforce that the function itself never gets called with a jwk containing a private key, and not have to do that here? if not might be good to add a unit test for this.

domleboss97 (approved):
💅🏻

packages/did/src/create-did-document.ts
venables: Great point. Turns out that method signature actually required it to be a PublicKeyJwk anyway, so I was able to drop this.

PR #88, opened 2026-05-15
fix(ack-pay): validate receipt payment option

EfeDurmaz16:
## Summary

- Validate that a PaymentReceipt's `paymentOptionId` matches an option offered by the verified Payment Request token
- Add a dedicated `InvalidPaymentReceiptError` for receipt-level semantic failures
- Cover the mismatched-option case with a regression test
- Add a patch changeset for `@agentcommercekit/ack-pay`

## Why

A receipt proves a `paymentRequestToken` and a selected `paymentOptionId`, but verification did not bind that selection back to an option actually offered by the request. This keeps receipt proof semantics aligned with the payment options the request offered.

The binding check reads from the proof-decoded credential (`verifiedReceipt`, returned by `verifyParsedCredential`), so it cannot be bypassed by mutating the outer object on the parsed-credential input path. This builds on the trust-boundary work in #108 / the proof-binding changeset already on `main`.

## Verification

- `pnpm install --frozen-lockfile`
- `pnpm run build`
- `pnpm run check` (build + format + types + lint + test) — all green, lint 0 warnings/0 errors
- `pnpm --filter @agentcommercekit/ack-pay test` — 35 passing

## AI Usage Disclosure

This contribution was AI-assisted (Codex CLI/app and Claude Opus). AI assistance was used for repository/docs navigation, understanding ACK context, and small edits/validation. The branch was rebased onto current `main` and the feature was rebuilt on top of main's proof-verified-receipt mechanism. I reviewed the final diff and take responsibility for the submitted changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Receipt verification now rejects receipts when the receipt’s `paymentOptionId` doesn’t match an option in the verified payment request.
  * Receipt parsing from JWT proof now derives receipt fields from the signed proof to prevent bypass via mutated input.
* **Tests**
  * Added a negative test ensuring `verifyPaymentReceipt(...)` throws when the payment option is missing from the payment request.
* **Chores**
  * Patch release update, including a new exported `InvalidPaymentReceiptError` for callers to catch receipt validation failures.

### DIFF
--- a/packages/ack-pay/src/verify-payment-receipt.ts
+++ b/packages/ack-pay/src/verify-payment-receipt.ts
@@ -10,6 +10,7 @@ import {
   type W3CCredential,
 } from "@agentcommercekit/vc"
 
+import { InvalidPaymentReceiptError } from "./errors"
 import type { PaymentRequest } from "./payment-request"
 import {
   getReceiptClaimVerifier,
@@ -127,6 +128,20 @@ export async function verifyPaymentReceipt(
       issuer: paymentRequestIssuer,
     },
   )
+
+  // Bind the receipt's selected option back to an option actually offered by
+  // the verified Payment Request. Reads from `verifiedReceipt` (proof-decoded),
+  // so a mutated outer credential cannot smuggle in an unoffered option.
+  const paymentOptionExists = paymentRequest.paymentOptions.some(
+    (paymentOption) =>
+      paymentOption.id === verifiedReceipt.credentialSubject.paymentOptionId,
+  )
+
+  if (!paymentOptionExists) {
+    throw new InvalidPaymentReceiptError(
+      "Receipt paymentOptionId does not match any payment option in the Payment Request token",
+    )
+  }
 
   return {
     receipt: verifiedReceipt,
--- a/.changeset/validate-receipt-payment-option.md
+++ b/.changeset/validate-receipt-payment-option.md
@@ -1,0 +1,16 @@
+---
+"@agentcommercekit/ack-pay": patch
+---
+
+Validate that a PaymentReceipt's `paymentOptionId` matches an option offered by
+the verified Payment Request token.
+
+A receipt proves a `paymentRequestToken` and a selected `paymentOptionId`, but
+verification did not bind that selection back to an option actually offered by
+the request. `verifyPaymentReceipt` now rejects a receipt whose
+`paymentOptionId` is not present in the verified request's `paymentOptions`. The
+check reads from the proof-decoded credential, so it cannot be bypassed by
+mutating the outer object on the parsed-credential input path.
+
+Introduces `InvalidPaymentReceiptError` so callers can catch receipt-level
+validation failures explicitly.
--- a/packages/ack-pay/src/errors.ts
+++ b/packages/ack-pay/src/errors.ts
@@ -7,3 +7,10 @@ export class InvalidPaymentRequestTokenError extends Error {
     this.name = "InvalidPaymentRequestTokenError"
   }
 }
+
+export class InvalidPaymentReceiptError extends Error {
+  constructor(message = "Invalid payment receipt", options?: ErrorOptions) {
+    super(message, options)
+    this.name = "InvalidPaymentReceiptError"
+  }
+}
--- a/packages/ack-pay/src/verify-payment-receipt.test.ts
+++ b/packages/ack-pay/src/verify-payment-receipt.test.ts
@@ -23,7 +23,10 @@ import { beforeEach, describe, expect, it } from "vitest"
 
 import { createPaymentReceipt } from "./create-payment-receipt"
 import { createSignedPaymentRequest } from "./create-signed-payment-request"
-import { InvalidPaymentRequestTokenError } from "./errors"
+import {
+  InvalidPaymentReceiptError,
+  InvalidPaymentRequestTokenError,
+} from "./errors"
 import type { PaymentRequestInit } from "./payment-request"
 import { isPaymentReceiptCredential } from "./receipt-claim-verifier"
 import { verifyPaymentReceipt } from "./verify-payment-receipt"
@@ -173,6 +176,27 @@ describe("verifyPaymentReceipt()", () => {
         metadata: evidenceMetadata,
       },
     })
+  })
+
+  it("throws when the receipt payment option is not in the payment request", async () => {
+    const mismatchedReceipt = createPaymentReceipt({
+      paymentRequestToken,
+      paymentOptionId: "missing-payment-option-id",
+      issuer: receiptIssuerDid,
+      payerDid: createDidPkhUri(
+        "eip155:84532",
+        "0x7B3D8F2E1C9A4B5D6E7F8A9B0C1D2E3F4A5B6C",
+      ),
+    })
+
+    const mismatchedReceiptJwt = await signCredential(mismatchedReceipt, {
+      did: receiptIssuerDid,
+      signer: createJwtSigner(receiptIssuerKeypair),
+    })
+
+    await expect(
+      verifyPaymentReceipt(mismatchedReceiptJwt, { resolver }),
+    ).rejects.toThrow(InvalidPaymentReceiptError)
   })
 
   it("throws for an invalid JWT receipt", async () => {


EfeDurmaz16:
Addressed the changeset nit in 5a4cb46 by mentioning the exported `InvalidPaymentReceiptError` for explicit receipt validation handling.\n\nVerification:\n- `git diff --check`\n- `pnpm exec oxfmt --check .changeset/validate-receipt-payment-option.md`

ak68a:
Hey @EfeDurmaz16, thanks for the burst of contributions here. 

Before we dig into individual PRs, one thing: our [AI Usage Policy](AI_POLICY.md) requires disclosure of AI tooling used in contributions, including the tool and the extent of assistance. Your PRs and issues don't include any disclosure, but we'd like to confirm: were any of these written or developed with AI assistance? If so, please update the PR descriptions accordingly. We hold all contributors to this standard, ourselves included.

Appreciate the initiative and will be reviewing each PR individually.

EfeDurmaz16:
Hey @ak68a ! 

Yes, I used AI assistance, mainly Codex CLI and the Codex app. The assistance was mostly for understanding ACK/Catena context, navigating docs/issues, identifying possible contribution areas, and helping draft or validate small changes. I reviewed the final diffs myself and can explain the changes and why they were made.

I should have disclosed this in the PR descriptions from the start. I’ll update the open PRs with the tooling and extent of AI assistance so they comply with the policy.

Appreciate the clarification, and thanks for taking the time to review them individually.

packages/ack-pay/src/verify-payment-receipt.ts
```diff
@@ -116,6 +117,18 @@ export async function verifyPaymentReceipt(
     },
   )
 
+  const receiptPaymentOptionId =
```
ak68a: Looks good. One note for future reference: this check runs against `parsedCredential.credentialSubject`, which is trustworthy when the input is a JWT string (decoded from the signed payload) but is the caller-supplied object when the input is a pre-parsed credential. 

The proof verification validates the JWT inside `proof`, but doesn't currently enforce that the outer object matches the signed payload. Not a blocker for this PR  tho.

ak68a:
Good fix, @EfeDurmaz16. A receipt could claim any `paymentOptionId` and verification wouldn't catch it. The implementation is clean and minimal.

One thing I want to flag. I dug into the security question re: parsed-credential path:

For **JWT inputs** (the primary path), `parseJwtCredential()` decodes from the signed payload, so `parsedCredential.credentialSubject.paymentOptionId` reflects the signed data.

For **pre-parsed credential inputs**, `verifyParsedCredential()` validates `proof.jwt`, but the new binding check reads from the outer `parsedCredential` object and not from the signed JWT payload directly. 

In theory, a caller could supply a valid receipt proof whose signed payload names one option but mutate the outer object to name another.

That said, this is a pre-existing trust assumption in the VC verification pipeline, not something introduced by your PR. Your change is strictly better than having no check at all. We can open a separate issue to track tightening the parsed-credential binding in `verifyParsedCredential` itself.

EfeDurmaz16:
Thanks for digging into that path.

Agreed, this sounds like a separate pre-existing trust boundary in the parsed credential verification flow rather than something this PR should expand into. I will keep this PR scoped to binding the receipt `paymentOptionId` against the verified payment request options.

I opened a focused follow-up issue for tightening `verifyParsedCredential()` so the outer parsed credential fields are bound back to the signed JWT payload: #105

venables:
@EfeDurmaz16 — there are two open inline review threads on `packages/ack-pay/src/verify-payment-receipt.ts` (lines 120 and 124), both raising the same point: the new `paymentOptionId` binding check reads from `parsedCredential.credentialSubject`, which is trustworthy when the input is a JWT string (decoded from the signed payload) but is caller-supplied when the input is a pre-parsed credential. The same applies to the existing `paymentRequestToken` read on that object, so it may be pre-existing rather than something this PR introduces.

What are your thoughts on these two comments? Do you see it as in-scope to address here (e.g. deriving the option from the signed proof), or better tracked as separate follow-up? The comments have no replies, so I'm curious what your take is.

EfeDurmaz16:
Thanks @venables — I went through the full path and I agree with both threads.

**Mechanism:** on the JWT-string input path, `parseJwtCredential` returns the credential reconstructed from the signed payload (`verifyCredential().verifiableCredential`), so `paymentOptionId` / `paymentRequestToken` / issuer are all trustworthy. On the parsed-credential path, `verifyParsedCredential` → `verifyProof` only verifies that `proof.jwt` is a validly-signed credential and then discards the decoded payload — every later check (claims, expiry, revocation, trusted-issuer) runs against the caller-supplied outer object. So a caller can attach any valid `proof.jwt` and set `credentialSubject.paymentOptionId` freely, and the binding check reads the spoofable field.

**Scope:** this is pre-existing rather than introduced here — the `paymentRequestToken` read just above has the identical exposure, and the `trustedReceiptIssuers` check compares the equally-caller-supplied outer `issuer.id` on this path. The new option-binding check inherits that trust model; on the JWT-string path (the documented usage) it does exactly what it should.

**What I changed here:** rather than leave the parsed path bypassable, I added a localized hardening — when the input is a parsed credential with a JWT proof, the receipt is now re-derived from `proof.jwt` (via the already-exported `parseJwtCredential`) before the `paymentOptionId` / `paymentRequestToken` reads, so they come from the signed payload. Added a regression test that a parsed credential whose outer `credentialSubject.paymentOptionId` is mutated to a valid option — while its signed proof names an option absent from the request — is now rejected (it fails without the change, passes with it). No public API change.

**Still recommend a follow-up:** the *general* fix belongs in `@agentcommercekit/vc` — after `verifyProof` validates `proof.jwt`, the credential fields should be derived from / reconciled against the decoded payload for **every** credential type (the `trustedIssuers` / claim checks share the same gap). That changes verification semantics library-wide, so I'd keep it as its own PR with dedicated tests rather than expand this one. Happy to open that issue and take it on.

Verification:
- `pnpm --filter @agentcommercekit/ack-pay test -- --run` (34 passing, incl. the new regression test)
- `pnpm --filter @agentcommercekit/ack-pay check:types`
- `pnpm exec oxfmt --check` / `oxlint` clean

EfeDurmaz16:
General `vc` fix tracked as a follow-up in #108.

packages/ack-pay/src/verify-payment-receipt.ts
```diff
@@ -68,7 +70,15 @@ export async function verifyPaymentReceipt(
...
   } else if (isCredential(receipt)) {
-    parsedCredential = receipt
+    // A parsed credential's top-level fields are caller-supplied and are not
+    // bound to the signed proof: verifyParsedCredential() verifies `proof.jwt`
+    // but does not reconcile the outer object against the decoded payload. When
+    // the proof is a JWT proof, re-derive the credential from `proof.jwt` so the
+    // paymentOptionId / paymentRequestToken reads below come from the signed
+    // payload rather than a (potentially mutated) caller-supplied object.
+    parsedCredential = isJwtProof(receipt.proof)
+      ? await parseJwtCredential(receipt.proof.jwt, resolver)
```
venables: Previously, when a `Verifiable<W3CCredential>` was passed in, `parsedCredential = receipt` and the proof was only validated later by `verifyParsedCredential()`, which throws `InvalidProofError` on a bad proof. Now `parseJwtCredential(receipt.proof.jwt, resolver)` runs first, so an invalid/malformed `proof.jwt` surfaces a raw `did-jwt-vc` error instead of `InvalidProofError` — a silent contract change for callers catching `InvalidProofError`.

Non-blocking. If you want to preserve the prior contract, re-throw as `InvalidProofError` around this call; otherwise it may be fine to leave as-is given the general `vc` trust-boundary work is already deferred to #108. Flagging so the choice is deliberate.

venables (dismissed):

venables:
Thanks @EfeDurmaz16 , this looks good to merge once the check/tests are passing

EfeDurmaz16:
@coderabbitai Thanks for the review. On the `isPaymentReceiptCredential(parsedCredential)` fast-reject (verify-payment-receipt.ts:77-83) — skipping this one, for the record:

- **Not introduced by this PR.** That precheck is existing `main` code. It only shows in this diff because the branch was rebased onto `main`, so CodeRabbit is diffing from the original PR base (`7d3eb3f`) and sees `main`'s intervening changes as new.
- **Intentional by design.** The maintainers added it with an explicit comment: *"Cheap structural fast-reject only — NOT a trust boundary. The authoritative receipt check runs on the proof-verified credential below."* The authoritative check is `isPaymentReceiptCredential(verifiedReceipt)` after `verifyParsedCredential()`.
- **Not a security issue.** The flagged behavior is a false-negative (rejecting a valid receipt whose outer object was mutated to not look like a receipt), not a false-accept. The trust boundary is intact.

This PR is scoped to binding the receipt's `paymentOptionId` back to an option offered by the verified Payment Request, on top of `main`'s proof-verified `verifiedReceipt`. Changing `main`'s deliberate fast-reject is out of scope here; happy to defer that to the maintainers if they want it gated to the JWT-string path.

packages/ack-pay/src/verify-payment-receipt.ts
```diff
@@ -116,6 +117,18 @@ export async function verifyPaymentReceipt(
     },
   )
 
+  const receiptPaymentOptionId =
+    parsedCredential.credentialSubject.paymentOptionId
+  const paymentOptionExists = paymentRequest.paymentOptions.some(
+    (paymentOption) => paymentOption.id === receiptPaymentOptionId,
+  )
```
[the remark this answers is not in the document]
EfeDurmaz16: Resolved by the rebase onto main: reads now come from the proof-decoded `verifiedReceipt` returned by `verifyParsedCredential`, so a mutated parsed credential can't smuggle in an unoffered option. This code path no longer exists.

packages/ack-pay/src/verify-payment-receipt.ts
EfeDurmaz16: Addressed after the rebase: the payment-option check reads from the proof-verified `verifiedReceipt`, not the caller-supplied object, so the trust-boundary concern no longer applies here.

packages/ack-pay/src/verify-payment-receipt.ts
EfeDurmaz16: No longer applicable — the `parseJwtCredential(receipt.proof.jwt)` re-parse this referred to was dropped in the rebase; the object path now relies on main's `verifiedReceipt` mechanism, so the `InvalidProofError` contract is unchanged.

venables (approved):
Ship it! 🚢

PR #96, opened 2026-05-15
feat(ack-pay): add approval model types

EfeDurmaz16:
## Summary
- Add minimal exported ACK-Pay types for approval requests and approval decisions
- Document how HITL approval objects fit around payment execution without adding a workflow engine to ACK core
- Clarify that denied decisions should stop before signing or settlement, while approved decisions can be recorded as audit evidence

Fixes #93

## Verification
- pnpm run build
- pnpm --filter ./packages/ack-pay check:types
- pnpm --filter ./packages/ack-pay exec vitest run
- pnpm exec oxfmt --check docs/ack-pay/hitl.mdx packages/ack-pay/src/index.ts packages/ack-pay/src/payment-approval.ts
- git diff --check

## AI Usage Disclosure
This contribution was AI-assisted using Codex CLI and the Codex app. AI assistance was used for repository/docs navigation, understanding ACK/Catena context, identifying relevant issues or contribution areas, and assisting with small edits/validation. I reviewed the final diff and take responsibility for the submitted changes.

## Summary by CodeRabbit

* **New Features**
  * Added payment approval request and decision types to ACK‑Pay to support approval workflows and decisions.

* **Documentation**
  * Added "Approval Request and Decision Shape" page with TypeScript examples illustrating request and decision structures.

* **Tests**
  * Added schema tests validating approval request and decision payloads across supported schema implementations.

* **Chores**
  * Included a changeset entry to publish a patch release with these additions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/agentcommercekit/ack/pull/96?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #104, opened 2026-05-25
refactor(ack-pay): extract timestamp schemas

EfeDurmaz16:
## Summary
- extract the repeated payment request timestamp normalization into local `timestampSchema` helpers
- reuse the helper from Valibot, Zod v3, and Zod v4 payment request schemas
- keep behavior unchanged and avoid adding any new public API surface

## Context
Split out from the review feedback on #96, where the timestamp schema extraction was identified as useful independently of the approval model proposal.

## Verification
- `corepack pnpm install --frozen-lockfile`
- `corepack pnpm --filter @agentcommercekit/ack-pay... build`
- `corepack pnpm --filter @agentcommercekit/ack-pay test -- --run`
- `corepack pnpm --filter @agentcommercekit/ack-pay check:types`
- `corepack pnpm exec oxfmt --check packages/ack-pay/src/schemas/valibot.ts packages/ack-pay/src/schemas/zod/v3.ts packages/ack-pay/src/schemas/zod/v4.ts`
- `git diff --check`

## Notes
- The first targeted test/typecheck run failed before building workspace dependencies because local package exports resolve from built `dist` outputs. Building the ACK-Pay dependency graph fixed that, and the rerun passed.
- No changeset is included because this is an internal refactor with no runtime behavior or exported API change.

## AI Usage Disclosure
This contribution was AI-assisted. Initial work used Codex CLI and the Codex app; the follow-up happy-path test was added with Claude Opus. AI assistance was used for repository/docs navigation, understanding ACK/Catena context, identifying relevant issues or contribution areas, and assisting with small edits/validation. I reviewed the final diff and take responsibility for the submitted changes.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated timestamp handling across schema implementations for consistency and reuse.

* **Bug Fix**
  * Validation now consistently rejects invalid timestamp inputs and normalizes valid ones to ISO timestamp strings.

* **Tests**
  * Added cross-schema tests covering invalid inputs and verifying valid Date/string inputs are normalized to the same ISO output.

### DIFF
--- a/packages/ack-pay/src/schemas.test.ts
+++ b/packages/ack-pay/src/schemas.test.ts
@@ -29,4 +29,23 @@ describe("paymentRequestSchema", () => {
     expect(zodV3PaymentRequestSchema.safeParse(input).success).toBe(false)
     expect(zodV4PaymentRequestSchema.safeParse(input).success).toBe(false)
   })
+
+  it("normalizes valid expiresAt inputs to an ISO string", () => {
+    const expected = "2024-12-31T23:59:59.000Z"
+    for (const expiresAt of [
+      new Date("2024-12-31T23:59:59Z"),
+      "2024-12-31T23:59:59Z",
+    ]) {
+      const input = { ...paymentRequest, expiresAt }
+
+      const valibot = v.safeParse(valibotPaymentRequestSchema, input)
+      expect(valibot.success && valibot.output.expiresAt).toBe(expected)
+
+      const v3 = zodV3PaymentRequestSchema.safeParse(input)
+      expect(v3.success && v3.data.expiresAt).toBe(expected)
+
+      const v4 = zodV4PaymentRequestSchema.safeParse(input)
+      expect(v4.success && v4.data.expiresAt).toBe(expected)
+    }
+  })
 })
--- a/packages/ack-pay/src/schemas.test.ts
+++ b/packages/ack-pay/src/schemas.test.ts
@@ -1,0 +1,32 @@
+import * as v from "valibot"
+import { describe, expect, it } from "vitest"
+
+import { paymentRequestSchema as valibotPaymentRequestSchema } from "./schemas/valibot"
+import { paymentRequestSchema as zodV3PaymentRequestSchema } from "./schemas/zod/v3"
+import { paymentRequestSchema as zodV4PaymentRequestSchema } from "./schemas/zod/v4"
+
+const paymentRequest = {
+  id: "test-payment-request-id",
+  paymentOptions: [
+    {
+      id: "test-payment-option-id",
+      amount: 10,
+      decimals: 2,
+      currency: "USD",
+      recipient: "sol:123",
+    },
+  ],
+}
+
+describe("paymentRequestSchema", () => {
+  it("reports invalid expiresAt strings as schema errors", () => {
+    const input = {
+      ...paymentRequest,
+      expiresAt: "invalid-date",
+    }
+
+    expect(v.safeParse(valibotPaymentRequestSchema, input).success).toBe(false)
+    expect(zodV3PaymentRequestSchema.safeParse(input).success).toBe(false)
+    expect(zodV4PaymentRequestSchema.safeParse(input).success).toBe(false)
+  })
+})
--- a/packages/ack-pay/src/schemas/valibot.ts
+++ b/packages/ack-pay/src/schemas/valibot.ts
@@ -6,6 +6,7 @@ const urlOrDidUri = v.union([v.pipe(v.string(), v.url()), didUriSchema])
 
 const timestampSchema = v.pipe(
   v.union([v.date(), v.string()]),
+  v.check((input) => !Number.isNaN(new Date(input).getTime()), "Invalid date"),
   v.transform((input) => new Date(input).toISOString()),
 )
 
--- a/packages/ack-pay/src/schemas/valibot.ts
+++ b/packages/ack-pay/src/schemas/valibot.ts
@@ -4,6 +4,11 @@ import * as v from "valibot"
 
 const urlOrDidUri = v.union([v.pipe(v.string(), v.url()), didUriSchema])
 
+const timestampSchema = v.pipe(
+  v.union([v.date(), v.string()]),
+  v.transform((input) => new Date(input).toISOString()),
+)
+
 export const paymentOptionSchema = v.object({
   id: v.string(),
   amount: v.union([v.pipe(v.number(), v.integer(), v.gtValue(0)), v.string()]),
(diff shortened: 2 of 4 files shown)


EfeDurmaz16:
Addressed the timestamp validation feedback in 224c426.

Changes:
- validate timestamp inputs before ISO normalization in Valibot, Zod v3, and Zod v4 schemas
- add a regression test covering invalid expiresAt strings across all three schema adapters

Verification:
- corepack pnpm --filter @agentcommercekit/ack-pay test -- --run
- corepack pnpm --filter @agentcommercekit/ack-pay check:types
- corepack pnpm exec oxfmt --check packages/ack-pay/src/schemas/valibot.ts packages/ack-pay/src/schemas/zod/v3.ts packages/ack-pay/src/schemas/zod/v4.ts packages/ack-pay/src/schemas.test.ts
- git diff --check

packages/ack-pay/src/schemas.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  it("reports invalid expiresAt strings as schema errors", () => {
+    const input = {
+      ...paymentRequest,
+      expiresAt: "invalid-date",
+    }
+
+    expect(v.safeParse(valibotPaymentRequestSchema, input).success).toBe(false)
+    expect(zodV3PaymentRequestSchema.safeParse(input).success).toBe(false)
+    expect(zodV4PaymentRequestSchema.safeParse(input).success).toBe(false)
+  })
```
venables, line 31: This only exercises the rejection path. Please add a happy-path test asserting that valid inputs are accepted **and normalized to an ISO string** across all three schemas — the `.toISOString()` transform is the load-bearing behavior here and is currently only covered indirectly (and only for valibot) by `create-signed-payment-request.test.ts`. A direct parity test guards against the three implementations drifting.

Suggested cases (a `Date` object and an equivalent string, both normalizing to the same ISO output):

```ts
it("normalizes valid expiresAt inputs to an ISO string", () => {
  const expected = "2024-12-31T23:59:59.000Z"
  for (const expiresAt of [
    new Date("2024-12-31T23:59:59Z"),
    "2024-12-31T23:59:59Z",
  ]) {
    const input = { ...paymentRequest, expiresAt }

    const valibot = v.safeParse(valibotPaymentRequestSchema, input)
    expect(valibot.success && valibot.output.expiresAt).toBe(expected)

    const v3 = zodV3PaymentRequestSchema.safeParse(input)
    expect(v3.success && v3.data.expiresAt).toBe(expected)

    const v4 = zodV4PaymentRequestSchema.safeParse(input)
    expect(v4.success && v4.data.expiresAt).toBe(expected)
  }
})
```

packages/ack-pay/src/schemas/valibot.ts
```diff
@@ -4,6 +4,12 @@ import * as v from "valibot"
 
 const urlOrDidUri = v.union([v.pipe(v.string(), v.url()), didUriSchema])
 
+const timestampSchema = v.pipe(
+  v.union([v.date(), v.string()]),
+  v.check((input) => !Number.isNaN(new Date(input).getTime()), "Invalid date"),
+  v.transform((input) => new Date(input).toISOString()),
+)
```
venables, line 11: Optional: Valibot ships an `isoTimestamp` action that can replace the hand-rolled NaN check here. The canonical pattern from valibot's own docs is `v.union([v.date(), v.pipe(v.string(), v.isoTimestamp())])`:

```ts
const timestampSchema = v.pipe(
  v.union([v.date(), v.pipe(v.string(), v.isoTimestamp())]),
  v.transform((input) => new Date(input).toISOString()),
)
```

Two upsides: it drops the custom `check`, and it removes the redundant double `new Date(input)` construction (currently built once in the `check` and again in the `transform`).

**Behavioral caveat to decide on first:** `isoTimestamp` only accepts strings that are already in ISO 8601 form, so loosely-formatted-but-parseable strings (e.g. `"2024/12/31"`, `"December 31 2024"`) that the current `new Date()`-based check accepts would now be rejected. For an `expiresAt` field that's arguably _more_ correct, but it is a tightening of the accepted input set. If you adopt it, mirror the change on the zod side (`z.string().datetime()` for v3, `z.iso.datetime()` for v4) so the three schemas stay in parity — otherwise the permissive `new Date()` version you have now is fine to keep.

venables (changes requested):
Thanks for tightening up the `expiresAt` handling — extracting the shared `timestampSchema` and rejecting invalid dates instead of letting `new Date("…").toISOString()` throw a `RangeError` is a solid fix.

Requesting one change before merge, plus an optional cleanup:

- **(please address)** The new `schemas.test.ts` only covers the rejection path. Add a happy-path test that asserts valid inputs are accepted _and normalized to an ISO string_ across all three schemas — that transform is the load-bearing behavior and it's only covered indirectly today. (Inline comment below.)
- **(optional)** Valibot has a built-in `isoTimestamp` action that can replace the hand-rolled NaN check and also remove the double `new Date()` construction. There's a behavioral caveat worth a conscious decision — details inline.

A panel of independent reviewers (claude-opus-4.6, qwen3.6-plus) otherwise found no correctness issues; claude empirically verified accept/reject parity across valibot / zod v3 / zod v4 for NaN, empty string, `Date` objects, negatives, fractionals, numeric strings, and `null`.

EfeDurmaz16:
Thanks @venables.

**Happy-path test (addressed):** added a direct parity test feeding both a `Date` object and the equivalent ISO string, asserting all three schemas accept them and normalize to the same ISO output (`2024-12-31T23:59:59.000Z`) — your suggested cases. This guards the `.toISOString()` transform directly across valibot / zod v3 / zod v4 rather than only indirectly via `create-signed-payment-request.test.ts`.

**isoTimestamp (kept current):** I went with the permissive `new Date()` check rather than `isoTimestamp` / `.datetime()`. The tightening is appealing, but `expiresAt` can arrive in loosely-formatted-but-valid forms and I'd rather not silently narrow the accepted set inside a refactor PR. If we want the stricter ISO-only contract I'll do it deliberately as its own change with matching `z.string().datetime()` / `z.iso.datetime()` so the three stay in parity — happy to follow up if you'd prefer that.

Verification:
- `pnpm --filter @agentcommercekit/ack-pay test -- --run` (34 passing)
- `pnpm --filter @agentcommercekit/ack-pay check:types`
- `pnpm exec oxfmt --check packages/ack-pay/src/schemas.test.ts`

venables:
Thanks for the clean refactor and the thorough verification notes, @EfeDurmaz16 — extracting the timestamp normalization was a sensible call.

Closing because this targets `packages/ack-pay/src/schemas/zod/v3.ts` and `zod/v4.ts`, both of which were removed in #116 (`feat!: adopt web-identity-schemas, drop zod v3`). The schema layout is now a single `schemas/zod.ts`, so the branch is conflicting and the v3 path no longer exists. The dedup idea is still reasonable against the new layout if you'd like to re-open a fresh PR scoped to `valibot.ts` + `zod.ts` — but no pressure, it's an internal refactor with no behavior change.

Appreciate the work here.

EfeDurmaz16:
Reopened this as #120, rescoped to the current `valibot.ts` + `zod.ts` layout (this branch targeted the pre-#116 `zod/v3` + `zod/v4` files, which no longer exist). Same fix — extract a shared `timestampSchema` that rejects unparseable `expiresAt` values instead of throwing a `RangeError` — plus the happy-path ISO-normalization parity test you asked for. `pnpm run check` is green.

Issue #108, opened 2026-06-04
vc: verifyParsedCredential trusts caller-supplied fields on parsed-credential input


EfeDurmaz16:
## Summary

`verifyParsedCredential()` verifies the JWT proof (`proof.jwt`) but does not reconcile the outer credential object against the decoded payload. On the **parsed-credential** input path, all subsequent checks therefore run against caller-supplied fields rather than the signed content.

## Where

- `packages/vc/src/verification/verify-proof.ts` — `verifyJwtProof()` calls `verifyCredential(proof.jwt, resolver)` and **discards the decoded result**; it only confirms the proof JWT is validly signed.
- `packages/vc/src/verification/verify-parsed-credential.ts` — expiry, revocation, the `trustedIssuers` check (reads `credential.issuer.id`), and every `ClaimVerifier` then operate on the **outer object** passed by the caller.

## Impact

When a caller passes a pre-parsed `Verifiable<W3CCredential>` (rather than a JWT string), they can attach *any* validly-signed `proof.jwt` and freely set the outer `credentialSubject`, `issuer`, expiry, etc. Concretely:

- Claim/subject fields are spoofable (this is the parsed-input gap raised in #88 for `paymentOptionId` / `paymentRequestToken`).
- The `trustedIssuers` check compares the caller-supplied `credential.issuer.id`, so it can be set to a trusted DID while `proof.jwt` is signed by an attacker's own (resolvable) DID.

The JWT-string input path is unaffected, since `parseJwtCredential()` reconstructs the credential from the signed payload.

## Proposed fix

After `verifyProof()` validates a JWT proof, derive the credential fields from the decoded `proof.jwt` payload (or reject parsed credentials whose outer object does not match the signed payload), so downstream checks operate on signed content for every credential type. This is cross-cutting, so it warrants its own change + tests rather than being handled per-consumer.

## Interim mitigation

#88 hardens the receipt path specifically by re-deriving receipt fields from `proof.jwt` when given a parsed credential. This issue tracks the general fix in `@agentcommercekit/vc`.

venables:
Fixed by #113 (merged): `verifyProof`/`verifyParsedCredential` now bind all trust decisions and returned values to the credential decoded from the verified proof, and `verifyPaymentReceipt` consumes that returned credential. Adversarial regression tests cover the parsed-object tampering path.

PR #110, opened 2026-06-13
fix(vc): bind credential verification to the signed proof (#105, #108)

### LINKED ISSUE #105: Tighten parsed credential binding to signed JWT payload
## Summary

`verifyParsedCredential()` verifies `credential.proof.jwt`, but downstream claim verifiers read fields from the outer parsed credential object. For pre-parsed credential inputs, that leaves a trust boundary where the signed JWT payload and the outer credential object can diverge.

## Context

This came up while reviewing #88. For JWT inputs, `parseJwtCredential()` decodes the credential from the signed payload, so downstream checks read signed data. For pre-parsed credential inputs, a caller can provide a valid proof JWT while mutating fields on the outer credential object before calling verification.

That is especially relevant for claim verifiers that enforce semantic bindings against `credentialSubject`, such as payment receipt validation.

## Expected behavior

When a credential includes `proof.jwt`, verification should ensure the outer credential fields used by downstream verifiers are bound to the signed JWT payload.

Possible approaches:

- Re-parse `proof.jwt` inside `verifyParsedCredential()` and compare the parsed credential body to the outer credential before running claim verifiers.
- Or normalize verification so claim verifiers receive the verified credential decoded from the JWT payload rather than the caller-provided outer object.

## Acceptance criteria

- Add a regression test where `proof.jwt` is valid but the outer `credentialSubject` is mutated.
- Ensure verification rejects the mismatch or runs claim verifiers against the signed payload.
- Keep the behavior for normal JWT-derived credentials unchanged.

## Notes

This is a pre-existing trust-boundary issue in parsed credential verification, not introduced by #88.

## AI Usage Disclosure

This issue was AI-assisted using Codex CLI and the Codex app. AI assistance was used for repository navigation, understanding the verification flow, and drafting the issue text. I reviewed the final text and take responsibility for the issue.

EfeDurmaz16:
## TL;DR

On the parsed-credential input path, `verifyParsedCredential()` verified `proof.jwt` but then ran expiry, revocation, `trustedIssuers`, and claim-verifier checks against the **caller-supplied** outer object. A caller could present a validly-signed `proof.jwt` while mutating the outer `credentialSubject` / `issuer` / `type`, and those mutated fields were trusted. This binds all verification to the signed payload. Fixes #105 and #108.

## What changed (`@agentcommercekit/vc`)

- `verifyProof()` now **returns the credential decoded from `proof.jwt`** (previously `void`).
- `verifyParsedCredential()` runs every downstream check against that verified credential, and **returns it**, so consumers can read signed fields from the return value instead of the object they passed in.

## Where to look

- `packages/vc/src/verification/verify-proof.ts` — `verifyJwtProof` returns the decoded credential.
- `packages/vc/src/verification/verify-parsed-credential.ts` — expiry / revocation / trusted-issuer / claim-verifier checks and the return value now use the verified credential.

## Risk / compatibility

- The return types of `verifyProof` and `verifyParsedCredential` change from `void` to the verified credential. This is backward-compatible: callers that ignore the return value are unaffected. Patch changeset included.
- The behavior change is intentional and scoped to the parsed-credential path: mutated outer fields no longer affect verification.

## Tests

- Rewrote `verify-parsed-credential.test.ts` onto real signed credentials (no mocked proof).
- Regression tests: a spoofed outer `issuer` (pointed at a trusted DID) is rejected; claim verifiers receive the signed subject rather than a mutated one; claim-verifier selection uses the signed `type`; and the returned credential is the signed one.
- `verify-proof.test.ts` asserts the decoded credential is returned.
- `@agentcommercekit/vc`: 51 tests passing. Repo-wide: `check:types`, `test`, `lint` (0 errors), and `oxfmt --check` all pass.

## Relationship to #88

#88 hardens the ACK-Pay receipt path specifically and adds the `paymentOptionId` ↔ Payment Request binding, which lives at the ACK-Pay layer and is not covered here. This PR is the general fix at the `@agentcommercekit/vc` layer and is complementary. With `verifyParsedCredential` now returning the canonical credential, the receipt path's local re-parse in #88 can later be simplified to consume the return value.

---

_Reviewed with ChatGPT 5.5 (xhigh reasoning) in addition to the author. Areas that may still warrant human review: (1) **consumer adoption** — `verifyParsedCredential` now returns the canonical credential, and call sites that read fields after verification should switch to the return value (the receipt path is separately hardened in #88); (2) **expiry / revocation binding** is covered structurally because those checks now receive the verified credential, but the unit tests mock `isExpired` / `isRevoked`, so a maintainer may want an end-to-end test with a genuinely expired or revoked signed credential._

## Summary by CodeRabbit

* **Bug Fixes**
  * Credential and proof verification is now bound to the signed proof payload, so spoofed or mutated outer fields are ignored.
  * `verifyProof` and `verifyParsedCredential` now return the verified/decoded credential for safer downstream handling.
  * Payment receipt verification now uses the verified receipt to prevent mismatched `paymentRequestToken` values.
* **Tests**
  * Strengthened verification test coverage with real signed JWT credentials and regression cases for spoofed issuer and altered claim fields.

### DIFF
--- a/packages/vc/src/verification/verify-proof.ts
+++ b/packages/vc/src/verification/verify-proof.ts
@@ -26,16 +26,24 @@ export function isJwtProof(proof: unknown): proof is JwtProof {
   )
 }
 
+/**
+ * Verify a JWT proof and return the credential decoded from the signed payload.
+ *
+ * The returned credential is reconstructed from `proof.jwt`, so its fields are
+ * exactly what was signed, rather than whatever a caller may have placed on the
+ * surrounding object.
+ */
 async function verifyJwtProof(
   proof: Verifiable<W3CCredential>["proof"],
   resolver: Resolvable,
-): Promise<void> {
+): Promise<Verifiable<W3CCredential>> {
   if (!isJwtProof(proof)) {
     throw new InvalidProofError()
   }
 
   try {
-    await verifyCredential(proof.jwt, resolver)
+    const { verifiableCredential } = await verifyCredential(proof.jwt, resolver)
+    return verifiableCredential as Verifiable<W3CCredential>
   } catch (_error) {
     throw new InvalidProofError()
   }
@@ -46,11 +54,14 @@ async function verifyJwtProof(
  *
  * @param proof - The credential proof to verify
  * @param resolver - The resolver to use for did resolution
+ * @returns The credential decoded from the signed proof. For JWT proofs this is
+ *   the payload recovered from `proof.jwt`, so callers can rely on it instead of
+ *   on caller-supplied top-level fields that are not bound to the signature.
  */
 export async function verifyProof(
   proof: Verifiable<W3CCredential>["proof"],
   resolver: Resolvable,
-): Promise<void> {
+): Promise<Verifiable<W3CCredential>> {
   switch (proof.type) {
     case "JwtProof2020":
       return verifyJwtProof(proof, resolver)
--- a/packages/vc/src/verification/verify-proof.ts
+++ b/packages/vc/src/verification/verify-proof.ts
@@ -9,6 +9,39 @@ interface JwtProof {
   jwt: string
 }
 
+type UnknownRecord = Record<string, unknown>
+
+function isRecord(value: unknown): value is UnknownRecord {
+  return typeof value === "object" && value !== null
+}
+
+function isStringArray(value: unknown): value is string[] {
+  return Array.isArray(value) && value.every((item) => typeof item === "string")
+}
+
+function isVerifiableCredential(
+  value: unknown,
+): value is Verifiable<W3CCredential> {
+  if (!isRecord(value)) {
+    return false
+  }
+
+  const issuer = value.issuer
+  const credentialSubject = value.credentialSubject
+  const proof = value.proof
+
+  return (
+    isStringArray(value["@context"]) &&
+    isStringArray(value.type) &&
+    typeof value.issuanceDate === "string" &&
+    isRecord(issuer) &&
+    typeof issuer.id === "string" &&
+    isRecord(credentialSubject) &&
+    isRecord(proof) &&
+    typeof proof.type === "string"
+  )
+}
+
 /**
  * Check if a proof is a JWT proof
  *
@@ -43,8 +76,16 @@ async function verifyJwtProof(
 
   try {
     const { verifiableCredential } = await verifyCredential(proof.jwt, resolver)
-    return verifiableCredential as Verifiable<W3CCredential>
+    if (!isVerifiableCredential(verifiableCredential)) {
+      throw new InvalidProofError("Verified credential has invalid shape")
+    }
+
+    return verifiableCredential
   } catch (_error) {
+    if (_error instanceof InvalidProofError) {
+      throw _error
+    }
+
     throw new InvalidProofError()
   }
 }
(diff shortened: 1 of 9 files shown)


packages/vc/src/verification/verify-proof.ts
```diff
@@ -26,16 +26,24 @@ export function isJwtProof(proof: unknown): proof is JwtProof {
...
-): Promise<void> {
+): Promise<Verifiable<W3CCredential>> {
   if (!isJwtProof(proof)) {
     throw new InvalidProofError()
   }
 
   try {
-    await verifyCredential(proof.jwt, resolver)
+    const { verifiableCredential } = await verifyCredential(proof.jwt, resolver)
+    return verifiableCredential as Verifiable<W3CCredential>
```
venables: Small / Optional polish: `verifiableCredential as Verifiable<W3CCredential>` is an unchecked type assertion. If `did-jwt-vc` ever returns a shape that diverges from `Verifiable<W3CCredential>`, the cast masks the mismatch.

**Possible Solution:** Validate the shape with a type guard or narrow the return type of `verifyCredential`.

venables:
**Location:** `demos/e2e/src/credential-verifier.ts:82-89` — https://github.com/agentcommercekit/ack/blob/794e138e17f797612a328c0537481665efc8e478/demos/e2e/src/credential-verifier.ts#L82-L89

After `verifyParsedCredential`, the demo returns the caller-supplied `parsedCredential` (whose outer fields are still unbound to the proof) instead of the now-returned verified credential, so downstream demo consumers read unsigned fields.

**Possible Solution:** Capture and return the value from `verifyParsedCredential` (`const verified = await verifyParsedCredential(...); return verified`).

venables:
**Location:** `examples/verifier/src/routes/verify.ts:44-50` — https://github.com/agentcommercekit/ack/blob/794e138e17f797612a328c0537481665efc8e478/examples/verifier/src/routes/verify.ts#L44-L50

Small / Optional polish: `verifyParsedCredential` return value is ignored. This endpoint only returns a success response (no credential fields leak), but the pattern demonstrates incorrect usage that downstream consumers may copy.

**Possible Solution:** Capture and use the return value, or add a comment documenting why the return value is intentionally discarded here.

venables (changes requested):
Thanks for putting this together — binding the downstream checks to the credential decoded from `proof.jwt` is the right fix and the new regression tests cover the tampering vectors well.

One thing to resolve before this lands: please check out the existing comment from the Codex reviewer on `packages/vc/src/verification/verify-parsed-credential.ts`. The core fix is correct, but the primary in-repo consumer — `verifyPaymentReceipt` in `packages/ack-pay/src/verify-payment-receipt.ts:82-90` — still ignores the new return value and reads `paymentRequestToken` from the caller-supplied object, so the #105/#108 tampering vector remains exploitable through the payment-receipt path. Capturing the returned verified credential and reading the receipt fields from it closes that gap.

I've also left a couple of smaller notes on the demo/example call sites and the unchecked cast.

venables:
Thanks for this @EfeDurmaz16 — solid work pinning down the root cause and tying it to #105/#108.

We're going to land this fix via #113, which takes the same approach (bind all trust decisions + returned values to the proof-decoded credential) and folds in the review feedback from here: it reuses `parseJwtCredential` instead of re-casting in `verify-proof.ts` (addressing the unchecked-cast comment), handles the receipt-path P1, and uses assertive test names. #113 also references #105/#108 so they'll close on merge.

Closing this in favor of #113 — appreciate you surfacing it.

PR #112, opened 2026-06-15
demo: wallet-state attestations (InsumerAPI) in ACK

douglasborthwick-crypto:
## What

Adds `demos/insumer-wallet-state`, mirroring `demos/skyfire-kya`: a worked example of consuming an InsumerAPI wallet-state attestation inside ACK.

## Why

ACK-ID attests *who* an agent is, and ACK-Pay confirms a payment cleared — but nothing checks the **paying wallet's on-chain state** before settlement: no balance, holdings, or compliance condition on the payer. InsumerAPI fills that gap: read a wallet against a condition (token balance / NFT / EAS) across 37 chains, return an ES256-signed boolean, verifiable offline against a public JWKS — the same envelope shape the `skyfire-kya` token rides.

## What it shows

- Verifies the attestation with the public JWKS (no secret) via the [`insumer-verify`](https://www.npmjs.com/package/insumer-verify) SDK — signature, condition-hash binding, expiry.
- Maps it into an ACK `WalletStateCredential` (W3C VC) with the original ES256 signature preserved as a detached JWS.
- Gates on the boolean, with InsumerAPI named in `trustedIssuers`. No re-signing, no fork.
- Runs against the live API (`pnpm demo`; a free key is one `curl` — see the README). In production today gating settlement for a Circle Alliance member building payments for the agentic economy.

The README documents a native path (publishing a `did:web` issuer + emitting VC-shaped output so it flows through `verifyParsedCredential`) as a follow-on, and is explicit about the bearer / presenter-binding boundary.

## AI usage disclosure

Built with AI assistance (Claude Code); reviewed and verified end-to-end against the live API, and type-checked in-tree.

## Summary by CodeRabbit

* **New Features**
  * Added a new `insumer-wallet-state` demo for consuming InsumerAPI wallet-state attestations to drive access-control gating.
  * Supports verifying attestations, handling expiry, and converting results into W3C Verifiable Credentials for ACK-style claim checking.
* **Documentation**
  * Added a README with two integration approaches, what the `granted` value means, and end-to-end run/mint/verify steps.
* **Tests / Tooling**
  * Added demo package setup (scripts/config) plus Vitest configuration and a sample JWT attestation for offline verification.


venables:
Thanks for the contribution, but closing this. The demo depends on a single-vendor SDK + live vendor API and reads as product promotion (InsumerAPI) rather than a vendor-neutral ACK demo. The bundled default-path sample attestation has also expired (2026-06-15), so the no-API-key path fails out of the box, and it re-introduces the `tsc --noEmit` script removed in #117. Happy to reconsider a self-contained, vendor-neutral wallet-state demo.

PR #116, opened 2026-06-23
feat!: adopt web-identity-schemas, drop zod v3

venables:
## Summary

Make [`web-identity-schemas`](https://github.com/catena-labs/web-identity-schemas) (v0.3.0) the source of truth for DID/JWT/VC validation schemas and DID/JWT types, replacing ACK's hand-rolled definitions, and **drop Zod v3 support** (w-i-s ships Zod v4 only).

- **Zod v3 dropped + exports collapsed.** Each package's `./schemas/zod/v3` + `./schemas/zod/v4` subpaths (and `./a2a/schemas/zod/*`) collapse into a single `./schemas/zod` (Zod v4). `zod` optional peer bumped to `^4.0.0`; catalog `zod` → `4.4.3`; `web-identity-schemas` added to the catalog.
- **DID** (`packages/did`): `didUriSchema` → w-i-s `DidSchema`; `isDidUri` delegates to w-i-s `isDid` (full DID-core validation); `DidUri` re-exported from w-i-s `Did`.
- **JWT** (`packages/jwt`): `jwtPayloadSchema` backed by w-i-s; `JwtString` re-exported from w-i-s. `jwtStringSchema` keeps ACK's non-empty-signature regex and `jwtHeaderSchema` keeps the narrow `ES256`/`ES256K`/`EdDSA` alg set.
- **VC** (`packages/vc`): `credentialSchema` backed by w-i-s' unsigned `CredentialV1Schema`, preserving ACK's `issuer` string→`{ id }` normalization, `type`/`@context` array coercion, and `JwtProof2020` passthrough. ACK's VC types (`W3CCredential`, `Verifiable`) stay local — `did-jwt-vc` remains the signing/verification engine.
- CAIP, payment, A2A, controller-claim, and `JwtProof2020` schemas remain hand-rolled (not covered by w-i-s).

## Breaking changes

- Dropped `./schemas/zod/v3` and `./schemas/zod/v4` exports → use `./schemas/zod`. `zod` peer is now `^4.0.0`.
- `didUriSchema`/`isDidUri` and `credentialSchema`/`isCredential` validate more strictly (DID-core syntax; VC Data Model v1.1 shape). ACK-issued credentials are unaffected; the credential-verification path (`parseJwtCredential`/`verifyParsedCredential`) is unchanged. See the changeset for details.

## Verification

`pnpm run check` is green (47/47 tasks: build + types + lint + format + test). `publint` clean for all published packages; `knip` clean. Includes a `major` changeset for the affected published packages.

[...]

PR #120, opened 2026-07-01
refactor(ack-pay): extract and harden expiresAt timestamp schema

EfeDurmaz16:
## Summary

Revives #104 (closed after #116 changed the schema layout), rescoped to the current `valibot.ts` + `zod.ts` layout — @venables invited a fresh PR against the new layout.

- Extract a shared `timestampSchema` for `paymentRequestSchema.expiresAt`
- **Reject unparseable dates** instead of throwing: previously `expiresAt` transformed via `new Date(input).toISOString()` directly, so a value like `"invalid-date"` threw a `RangeError` out of `safeParse` rather than producing a validation error
- Keep valibot and zod in accept/reject parity
- Add `schemas.test.ts` covering both the rejection path and the happy-path normalization to ISO across both schemas (the load-bearing transform, per @venables' request on #104)

## Why

`new Date("invalid").toISOString()` throws a `RangeError`. Callers using `safeParse` expect a `{ success: false }` result, not an exception. The shared schema validates the date is parseable before normalizing.

## Verification

- `pnpm run build`
- `pnpm run check` (build + format + type-aware lint + typecheck + test) — all green, lint 0/0
- `pnpm --filter @agentcommercekit/ack-pay test` — 36 passing (incl. new `schemas.test.ts`)

## Notes

Kept the permissive `new Date()` parse check rather than valibot's `isoTimestamp` / zod's `.datetime()` — narrowing the accepted `expiresAt` set to ISO-only is a deliberate contract change and out of scope for this dedup, same rationale as on #104. Happy to follow up with a stricter, parity-matched version if preferred.

## AI Usage Disclosure

This contribution was AI-assisted using Claude Code (Claude Opus). AI assistance was used to port the original #104 change onto the new schema layout, adapt the test to the two-schema (valibot + zod) structure, and run verification (`pnpm run check`). I reviewed the final diff and understand what it does — it validates `expiresAt` is a parseable date before normalizing to an ISO string, keeping valibot and zod in parity — and take responsibility for the submitted changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid `expiresAt` values now fail validation cleanly instead of causing runtime errors.
  * Date inputs are now handled consistently, with valid values converted to ISO timestamps and invalid ones rejected.
* **Tests**
  * Added coverage for both schema implementations to verify invalid dates are rejected and valid dates are normalized correctly.
* **Documentation**
  * Added a release note entry describing the schema behavior update.

### DIFF
--- a/.changeset/dedupe-timestamp-schema.md
+++ b/.changeset/dedupe-timestamp-schema.md
@@ -1,0 +1,12 @@
+---
+"@agentcommercekit/ack-pay": patch
+---
+
+Reject invalid `expiresAt` values in `paymentRequestSchema` instead of throwing.
+
+Previously the `expiresAt` transform called `new Date(input).toISOString()`
+directly, so an unparseable value (e.g. `"invalid-date"`) threw a `RangeError`
+out of `safeParse` rather than producing a validation error. A shared
+`timestampSchema` now validates the date is parseable before normalizing it to
+an ISO string, and the valibot and zod schemas share the same accept/reject
+behavior.
--- a/packages/ack-pay/src/schemas.test.ts
+++ b/packages/ack-pay/src/schemas.test.ts
@@ -1,0 +1,46 @@
+import * as v from "valibot"
+import { describe, expect, it } from "vitest"
+
+import { paymentRequestSchema as valibotPaymentRequestSchema } from "./schemas/valibot"
+import { paymentRequestSchema as zodPaymentRequestSchema } from "./schemas/zod"
+
+const paymentRequest = {
+  id: "test-payment-request-id",
+  paymentOptions: [
+    {
+      id: "test-payment-option-id",
+      amount: 10,
+      decimals: 2,
+      currency: "USD",
+      recipient: "sol:123",
+    },
+  ],
+}
+
+describe("paymentRequestSchema", () => {
+  it("rejects invalid expiresAt strings instead of throwing", () => {
+    const input = {
+      ...paymentRequest,
+      expiresAt: "invalid-date",
+    }
+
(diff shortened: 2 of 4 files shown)


.changeset/dedupe-timestamp-schema.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Reject invalid `expiresAt` values in `paymentRequestSchema` instead of throwing.
+
+Previously the `expiresAt` transform called `new Date(input).toISOString()`
+directly, so an unparseable value (e.g. `"invalid-date"`) threw a `RangeError`
+out of `safeParse` rather than producing a validation error. A shared
+`timestampSchema` now validates the date is parseable before normalizing it to
+an ISO string, and the valibot and zod schemas share the same accept/reject
+behavior.
```
venables, line 12: Small / Optional polish: The changeset names `timestampSchema`, but the schema is not exported. Consumers read this changelog and can look for an export that does not exist.

```suggestion
out of `safeParse` rather than producing a validation error. The schema now
validates that the date is parseable before it normalizes the value to an ISO
string, and the valibot and zod schemas share the same accept/reject behavior.
```

packages/ack-pay/src/schemas.test.ts
```diff
@@ -0,0 +1,46 @@
+import * as v from "valibot"
```
venables, line 1: Small / Optional polish: The test file sits in `src/`, but the code it tests sits in `src/schemas/`. The repo rule says tests are co-located with source.

**Possible Solution:** Move the file to `packages/ack-pay/src/schemas/schemas.test.ts` and change the imports to `./valibot` and `./zod`.

venables (approved):
Ship it! 🚢

PR #129, opened 2026-07-22
feat(did): add optional fetch timeout to the did:web resolver

EfeDurmaz16:
## What
Adds an optional `timeout` (ms) to `DidWebResolverOptions` / `getResolver` for
the `did:web` resolver. When set, the resolver passes
`AbortSignal.timeout(timeout)` to the underlying fetch.

## Why
Resolving a `did:web` DID fetches the host named in the DID. Today
`fetchDidDocumentAtUrl` calls `fetch(url, { mode: "cors" })` with no
`AbortSignal`, so an unresponsive or slow host hangs the resolver
indefinitely. The resolver's own header comment says it exists for "additional
checks and more control for fetching and resolution", so a bounded fetch fits
its purpose. This is resource-exhaustion hardening, not an SSRF fix (did:web
inherently fetches the domain named in the DID).

## Change
- `timeout?: number` on `DidWebResolverOptions`
- threaded through `getResolver` -> `resolve` -> `fetchDidDocumentAtUrl`
- omitted = no timeout, so existing behaviour is unchanged (backward compatible)

## Tests
Three new tests in `web-did-resolver.test.ts`: the resolver builds the signal
from the configured timeout (`AbortSignal.timeout` is called with the value),
passes no signal when the option is omitted, and surfaces an aborted/timed-out
fetch as the standard `notFound` resolution error.
`pnpm --filter @agentcommercekit/did test` -> 73 passing. Changeset included
(minor). A response-size cap is a natural follow-up but needs body streaming,
so it is out of scope here.

## AI assistance disclosure
Per the repository AI policy: implemented and tested by gpt 5.6-sol and fable 5, then reviewed and understood by me. I can explain the implementation and its trade-offs without AI aid. Tests, typecheck, and formatting pass locally against the package suite.

## Summary by CodeRabbit

* **New Features**
  * Added an optional `timeout` setting for `did:web` resolution requests (defaults to 5000ms).
  * Resolver requests are automatically cancelled when the timeout elapses (by wiring an `AbortSignal` to the fetch call).
* **Bug Fixes**
  * Invalid `timeout` values now fail fast with a clear `RangeError` (valid range: `1..2147483647`).
  * Timeout failures surface as a `notFound` resolution result with timeout-related error details.
* **Documentation**
  * Updated resolver docs to clarify timeout behavior and limitations.
* **Tests**
  * Updated/added unit tests covering signal wiring, default timeout, invalid values, and timeout error mapping.

### DIFF
--- a/packages/did/src/did-resolvers/web-did-resolver.test.ts
+++ b/packages/did/src/did-resolvers/web-did-resolver.test.ts
@@ -60,7 +60,7 @@ describe("web-did-resolver", () => {
       })
       expect(mockFetch).toHaveBeenCalledWith(
         "https://example.com/.well-known/did.json",
-        { mode: "cors" },
+        { mode: "cors", signal: expect.any(AbortSignal) },
       )
     })
 
@@ -92,7 +92,7 @@ describe("web-did-resolver", () => {
 
       expect(mockFetch).toHaveBeenCalledWith(
         "https://example.com/custom/path/did.json",
-        { mode: "cors" },
+        { mode: "cors", signal: expect.any(AbortSignal) },
       )
     })
 
@@ -125,7 +125,7 @@ describe("web-did-resolver", () => {
 
       expect(mockFetch).toHaveBeenCalledWith(
         "http://localhost:8787/.well-known/did.json",
-        { mode: "cors" },
+        { mode: "cors", signal: expect.any(AbortSignal) },
       )
     })
 
@@ -161,7 +161,7 @@ describe("web-did-resolver", () => {
 
       expect(mockFetch).toHaveBeenCalledWith(
         "https://example.com/issuers/v1/did.json",
-        { mode: "cors" },
+        { mode: "cors", signal: expect.any(AbortSignal) },
       )
     })
 
@@ -197,7 +197,7 @@ describe("web-did-resolver", () => {
 
       expect(mockFetch).toHaveBeenCalledWith(
         "http://localhost:8787/issuers/v1/did.json",
-        { mode: "cors" },
+        { mode: "cors", signal: expect.any(AbortSignal) },
       )
     })
 
@@ -414,11 +414,12 @@ describe("web-did-resolver", () => {
       timeoutSpy.mockRestore()
     })
 
-    it("does not pass a signal when no timeout is set", async () => {
+    it("applies the 5000ms default timeout when none is set", async () => {
       mockFetch.mockResolvedValueOnce({
         ok: true,
         json: () => Promise.resolve(mockDidDocument),
       })
+      const timeoutSpy = vi.spyOn(AbortSignal, "timeout")
 
       const did = "did:web:example.com"
       const resolver = getResolver()
@@ -440,18 +441,22 @@ describe("web-did-resolver", () => {
         {},
       )
 
-      // Exact init, so this fails if a signal (or anything else) is added.
+      expect(timeoutSpy).toHaveBeenCalledWith(5000)
       expect(mockFetch).toHaveBeenCalledWith(
         "https://example.com/.well-known/did.json",
-        { mode: "cors" },
+        expect.objectContaining({
+          mode: "cors",
+          signal: expect.any(AbortSignal),
+        }),
       )
+      timeoutSpy.mockRestore()
     })
 
     it("throws for an invalid timeout", () => {
-      expect(() => getResolver({ timeout: 0 })).toThrow(TypeError)
-      expect(() => getResolver({ timeout: -1 })).toThrow(TypeError)
-      expect(() => getResolver({ timeout: 1.5 })).toThrow(TypeError)
-      expect(() => getResolver({ timeout: NaN })).toThrow(TypeError)
+      expect(() => getResolver({ timeout: 0 })).toThrow(RangeError)
+      expect(() => getResolver({ timeout: -1 })).toThrow(RangeError)
+      expect(() => getResolver({ timeout: 1.5 })).toThrow(RangeError)
+      expect(() => getResolver({ timeout: NaN })).toThrow(RangeError)
     })
 
     it("surfaces a timed-out fetch as a notFound resolution error", async () => {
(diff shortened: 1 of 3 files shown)


packages/did/src/did-resolvers/web-did-resolver.ts
```diff
@@ -57,9 +67,15 @@ const DEFAULT_DOC_PATH = "/.well-known/did.json"
...
+  {
+    fetch = globalThis.fetch,
+    timeout,
+  }: { fetch?: FetchLike; timeout?: number } = {},
 ): Promise<DidDocument> {
-  const res = await fetch(url, { mode: "cors" })
+  const res = await fetch(url, {
+    mode: "cors",
+    ...(timeout !== undefined ? { signal: AbortSignal.timeout(timeout) } : {}),
+  })
```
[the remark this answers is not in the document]
EfeDurmaz16: Fixed in f288d45: `getResolver` now throws a TypeError for a non-positive, non-integer or non-finite `timeout` (fail fast instead of surfacing a misleading `notFound`), with tests for 0 / -1 / 1.5 / NaN. Thanks.

EfeDurmaz16:
@venables when you have a moment, would appreciate a review. Small backward-compatible hardening (optional fetch timeout on the did:web resolver) I hit while building an ACK-ID + x402 demo. CodeRabbit's note is addressed; CI is waiting on a maintainer 'approve and run' for the fork branch.

packages/did/src/did-resolvers/web-did-resolver.test.ts
```diff
@@ -375,5 +375,115 @@ describe("web-did-resolver", () => {
       expect(customFetch).toHaveBeenCalled()
       expect(mockFetch).not.toHaveBeenCalled()
     })
+
+    it("passes an abort signal built from the configured timeout", async () => {
```
domleboss97, line 379: not related to this pr in particular, but i think some test helpers could make each of these tests less verbose

packages/did/src/did-resolvers/web-did-resolver.ts
```diff
@@ -44,6 +44,16 @@ export interface DidWebResolverOptions {
    * @default []
    */
   allowedHttpHosts?: string[]
+  /**
```
domleboss97, line 47: this comment seems a bit verbose - it seems like we could drop the rationale sentence and the mention of previous behavior, which one could see through changesets anyways. could also add a `@default` annotation

packages/did/src/did-resolvers/web-did-resolver.ts
```diff
@@ -141,7 +157,15 @@ export function getResolver({
   docPath = DEFAULT_DOC_PATH,
   fetch = globalThis.fetch,
   allowedHttpHosts = DEFAULT_ALLOWED_HTTP_HOSTS,
+  timeout,
```
domleboss97: might make sense now that we're adding this to have a default value, perhaps 5 seconds or something?

packages/did/src/did-resolvers/web-did-resolver.ts
```diff
@@ -141,7 +157,15 @@ export function getResolver({
...
 }: DidWebResolverOptions = {}): { web: DIDResolver } {
+  // Fail fast on a bad timeout: `AbortSignal.timeout` throws on non-positive,
```
domleboss97: very minor - `AbortSignal.timeout(0)` does not actually throw

packages/did/src/did-resolvers/web-did-resolver.ts
```diff
@@ -141,7 +157,15 @@ export function getResolver({
...
+  // non-integer or non-finite values, and that would otherwise surface as a
+  // misleading `notFound` resolution error rather than a programmer error.
+  if (timeout !== undefined && (!Number.isInteger(timeout) || timeout <= 0)) {
+    throw new TypeError("`timeout` must be a positive integer (milliseconds)")
```
domleboss97: nit - `RangeError` might be more accurate here. the type is actually correct.

domleboss97 (dismissed):
😎 some comments, but looks good!

packages/did/src/did-resolvers/web-did-resolver.ts
EfeDurmaz16: Trimmed and added @default 5000 in 814f8d8.

packages/did/src/did-resolvers/web-did-resolver.ts
EfeDurmaz16: Agreed, switched in 814f8d8, tests updated.

packages/did/src/did-resolvers/web-did-resolver.test.ts
EfeDurmaz16: Agreed it would help. Left it out since it touches the pre-existing tests too; happy to do a small follow-up PR.

packages/did/src/did-resolvers/web-did-resolver.ts
```diff
@@ -44,6 +44,14 @@ export interface DidWebResolverOptions {
...
    */
   allowedHttpHosts?: string[]
+  /**
+   * Milliseconds to wait for the DID document fetch before aborting.
+   *
+   * The timeout is applied via an `AbortSignal` on the request. A custom
+   * `fetch` must honour `init.signal` for it to take effect.
+   * @default 5000
+   */
+  timeout?: number
```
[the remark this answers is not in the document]
EfeDurmaz16: Intentional: the default was requested in review, see the thread above. The changeset documents it as a behavior change, including the accepted range and the effective opt-out at the max value. Keeping it.

packages/did/src/did-resolvers/web-did-resolver.ts
EfeDurmaz16: Done in 814f8d8: defaults to 5000ms now. While double checking the range for this I noticed values above 2^31-1 clamp to a 1ms timer in Node, and above 2^32 they throw inside the fetch, both surfacing as a misleading notFound. 

So f1e9670 caps the range at 2^31 -1 . Passing the max (about 24.8 days) works as an effective opt-out. The changeset covers both.

packages/did/src/did-resolvers/web-did-resolver.ts
EfeDurmaz16: Right, it doesn't throw natively. getResolver still rejects 0 with a RangeError since a 0ms timeout would abort every request before it starts. Fixed the comment in 814f8d8 so it no longer claims the API throws on it.

domleboss97 (approved):
🚀

PR #133, opened 2026-07-28
feat(did): refuse redirects when resolving did:web by default

EfeDurmaz16:
## What

`getResolver` for `did:web` now sends `redirect: "error"` when fetching the DID document. A new `followRedirects?: boolean` option (default `false`) restores the previous behavior.

## Why

`allowedHttpHosts` is applied to the URL built from the DID:

```ts
const url = isHttpAllowed(path, allowedHttpHosts)
  ? `http://${path}`
  : `https://${path}`
```

The fetch that follows it used the platform default, which follows redirects. So the allowlist only governs the first hop: a `did:web` resolved over `https` can answer `302 Location: http://…`, and the resolver follows it to a scheme and host the check was there to reject.

```mermaid
flowchart LR
  D["did:web:issuer.example"] --> C{"allowedHttpHosts<br/>allows http for this host?"}
  C -- no --> U["fetch https://issuer.example/.well-known/did.json"]
  U --> H{"host answers 302<br/>Location: http://internal/…"}
  H -->|before: redirect followed| BAD["request lands on the scheme<br/>and host the check rejected"]
  H -->|after: redirect error| OK["resolution fails as notFound"]

  classDef bad stroke-dasharray: 4 3
  class BAD bad
```

Because the redirect target is chosen by the DID's own host, this also makes resolution an outbound request an untrusted party controls. That matters for anything that resolves a DID *before* verifying a signature, which is the normal order: you need the document to check the signature. A server verifying an ACK-ID proof from an arbitrary issuer will fetch whatever that issuer's host points it at.

did:web documents are served directly at `/.well-known/did.json` (or the configured `docPath`), so refusing redirects costs legitimate resolution nothing. Deployments that genuinely sit behind a redirect can opt in.

## How

- `DidWebResolverOptions` gains `followRedirects?: boolean`, documented with the reason, defaulting to `false`.
- `fetchDidDocumentAtUrl` passes `redirect: followRedirects ? "follow" : "error"` alongside the existing `mode: "cors"`.
- Existing tests that assert the fetch options were updated to include the new field; two tests cover the default (refuses, resolves to `notFound`) and the opt-in (`redirect: "follow"`).

## Test plan

```sh
pnpm build
pnpm --filter @agentcommercekit/did test   # 72 passed
```

Found while building an ACK-ID identity-gated x402 demo, where a seller resolves buyer DIDs it has never seen before.

Implemented and tested by gpt 5.6-sol and fable 5, reviewed by me.

## Summary by CodeRabbit

* **New Features**
  * Added optional `followRedirects: true` support for `did:web` and `did:jwks` resolution.
  * Redirects are now blocked by default for safer resolution.
  * Blocked redirects report the destination when available; otherwise, resolution returns a clear not-found or resolution error.
  * Enabling redirect following preserves existing successful resolution behavior.

* **Documentation**
  * Documented redirect-handling behavior and the `followRedirects` option.

### DIFF
--- a/packages/did/src/did-resolvers/web-did-resolver.ts
+++ b/packages/did/src/did-resolvers/web-did-resolver.ts
@@ -44,6 +44,17 @@ export interface DidWebResolverOptions {
    * @default []
    */
   allowedHttpHosts?: string[]
+  /**
+   * Whether to follow HTTP redirects while fetching the did document.
+   *
+   * The `allowedHttpHosts` check applies to the resolved URL only, so a
+   * followed redirect can move the request to a host or scheme that check
+   * would have rejected. did:web documents are served directly at a
+   * well-known path, so redirects are refused by default.
+   *
+   * @default false
+   */
+  followRedirects?: boolean
 }
 
 const DEFAULT_ALLOWED_HTTP_HOSTS: string[] = []
@@ -57,9 +68,15 @@ const DEFAULT_DOC_PATH = "/.well-known/did.json"
  */
 async function fetchDidDocumentAtUrl(
   url: string | URL,
-  { fetch = globalThis.fetch }: { fetch?: FetchLike } = {},
+  {
+    fetch = globalThis.fetch,
+    followRedirects = false,
+  }: { fetch?: FetchLike; followRedirects?: boolean } = {},
 ): Promise<DidDocument> {
-  const res = await fetch(url, { mode: "cors" })
+  const res = await fetch(url, {
+    mode: "cors",
+    redirect: followRedirects ? "follow" : "error",
+  })
 
   if (!res.ok) {
     throw new Error(
@@ -141,6 +158,7 @@ export function getResolver({
   docPath = DEFAULT_DOC_PATH,
   fetch = globalThis.fetch,
   allowedHttpHosts = DEFAULT_ALLOWED_HTTP_HOSTS,
+  followRedirects = false,
 }: DidWebResolverOptions = {}): { web: DIDResolver } {
   async function resolve(
     did: string,
@@ -155,7 +173,10 @@ export function getResolver({
     let didDocument: DIDDocument | null = null
 
     try {
-      didDocument = await fetchDidDocumentAtUrl(url, { fetch })
+      didDocument = await fetchDidDocumentAtUrl(url, {
+        fetch,
+        followRedirects,
+      })
 
       if (!isDidDocumentForDid(didDocument, did)) {
         throw new Error("DID document id does not match requested did")
--- a/packages/did/src/did-resolvers/web-did-resolver.ts
+++ b/packages/did/src/did-resolvers/web-did-resolver.ts
@@ -75,8 +75,21 @@ async function fetchDidDocumentAtUrl(
 ): Promise<DidDocument> {
   const res = await fetch(url, {
     mode: "cors",
-    redirect: followRedirects ? "follow" : "error",
+    redirect: followRedirects ? "follow" : "manual",
   })
+
+  // With `redirect: "manual"` a redirect resolves instead of throwing, so we
+  // can report it precisely. Node exposes the 3xx status and Location header;
+  // browsers return an opaque redirect (type "opaqueredirect", status 0).
+  if (
+    !followRedirects &&
+    (res.type === "opaqueredirect" || (res.status >= 300 && res.status < 400))
+  ) {
+    const location = res.headers.get("location")
+    throw new Error(
+      `DID resolution refused a redirect${location ? ` to ${location}` : ""}`,
+    )
+  }
 
   if (!res.ok) {
     throw new Error(
--- a/packages/did/src/did-resolvers/web-did-resolver.ts
+++ b/packages/did/src/did-resolvers/web-did-resolver.ts
@@ -52,6 +52,8 @@ export interface DidWebResolverOptions {
    * would have rejected. did:web documents are served directly at a
    * well-known path, so redirects are refused by default.
    *
+   * The policy is applied via `init.redirect` on the request. A custom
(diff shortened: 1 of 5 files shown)


packages/did/src/did-resolvers/web-did-resolver.ts
```diff
@@ -57,9 +68,15 @@ const DEFAULT_DOC_PATH = "/.well-known/did.json"
...
-  { fetch = globalThis.fetch }: { fetch?: FetchLike } = {},
+  {
+    fetch = globalThis.fetch,
+    followRedirects = false,
+  }: { fetch?: FetchLike; followRedirects?: boolean } = {},
 ): Promise<DidDocument> {
-  const res = await fetch(url, { mode: "cors" })
+  const res = await fetch(url, {
+    mode: "cors",
+    redirect: followRedirects ? "follow" : "error",
```
domleboss97: i think it might be more useful to do `manual` here. with `error` this will throw with a `TypeError`, and then I think the resolver has nothing better ot report than `notFound`.

with `manual`, fetch resolves with the redirect repsonse (although in browsers i think it's opaque response), so we could throw a precise error - on node it could include the redirect target from the `Location` header.

domleboss97 (dismissed):
One comment on the error handling, but I think this is the right move!

packages/did/src/did-resolvers/web-did-resolver.ts
EfeDurmaz16: Done in 22f6a5d: switched to redirect: "manual". The resolver now refuses any redirect response with an error naming the Location target on Node; browsers surface an opaque redirect so the same error fires without a target. Tests cover both cases, changeset updated.

packages/did/src/did-resolvers/web-did-resolver.ts
```diff
@@ -57,9 +68,28 @@ const DEFAULT_DOC_PATH = "/.well-known/did.json"
...
+  // With `redirect: "manual"` a redirect resolves instead of throwing, so we
+  // can report it precisely. Node exposes the 3xx status and Location header;
+  // browsers return an opaque redirect (type "opaqueredirect", status 0).
+  if (
+    !followRedirects &&
+    (res.type === "opaqueredirect" || (res.status >= 300 && res.status < 400))
+  ) {
+    const location = res.headers.get("location")
+    throw new Error(
+      `DID resolution refused a redirect${location ? ` to ${location}` : ""}`,
```
domleboss97: nit - might be nice to add an action here, i.e. ` Set followRedirects: true to allow redirects.`

domleboss97 (dismissed):
🚀

domleboss97 (dismissed):

venables:
@EfeDurmaz16 mind resolving the conflicts?

EfeDurmaz16:
@venables resolved!

packages/did/src/did-resolvers/web-did-resolver.ts
```diff
@@ -44,6 +44,17 @@ export interface DidWebResolverOptions {
...
+   * Whether to follow HTTP redirects while fetching the did document.
+   *
+   * The `allowedHttpHosts` check applies to the resolved URL only, so a
+   * followed redirect can move the request to a host or scheme that check
+   * would have rejected. did:web documents are served directly at a
+   * well-known path, so redirects are refused by default.
+   *
+   * @default false
+   */
+  followRedirects?: boolean
```
venables, line 59: Small / Optional polish: The `followRedirects` doc does not say that a custom `fetch` must honour `init.redirect`. A custom `fetch` that ignores it silently follows redirects, and the 3xx check at line 100-103 never fires. The `timeout` option (line 62-63) already carries the equivalent note for `init.signal`.

```suggestion
  /**
   * Whether to follow HTTP redirects while fetching the did document.
   *
   * The `allowedHttpHosts` check applies to the resolved URL only, so a
   * followed redirect can move the request to a host or scheme that check
   * would have rejected. did:web documents are served directly at a
   * well-known path, so redirects are refused by default.
   *
   * A custom `fetch` must honour `init.redirect` for this to take effect.
   *
   * @default false
   */
```

venables (changes requested):
**Location:** `packages/did/src/did-resolvers/get-did-resolver.ts:34-37` — https://github.com/agentcommercekit/ack/blob/b3be83c1e7f0fe25ef931be5f8984fc626bcd094/packages/did/src/did-resolvers/get-did-resolver.ts#L34-L37

`getDidResolver` spreads the same `webOptions` into `jwks-did-resolver`, but `did-jwks` calls `fetchImpl(url)` with no `init` (see `node_modules/did-jwks/dist/fetch-*.js:125`), so the `did:jwks` path still follows redirects after this change. A caller who sets `webOptions.followRedirects` (or relies on the new default) gets protection for `did:web` only, and the same `allowedHttpHosts` bypass remains for `did:jwks`.

**Possible Solution:** in the existing fetch wrapper at line 36, pass `{ ...init, redirect: webOptions.followRedirects ? "follow" : "manual" }` (`did-jwks` treats a 3xx as `!resp.ok` and returns null), or state in the `followRedirects` docs that the option applies to `did:web` only.

EfeDurmaz16:
Addressed in c0279ce: the redirect policy is now injected into the fetch passed to jwks-did-resolver, since did-jwks calls fetch with no init. A refused redirect fails the ok check and did:jwks resolution fails as notFound. Added tests for the default, followRedirects: true, and the global-fetch path, and updated the changeset. The followRedirects doc note about custom fetch honouring init.redirect landed in e4efdce.

venables (approved):

venables:
Thank you for the thorough work on this one, including the jwks resolver follow-up and the redirect tests. Merging.

PR #137, opened 2026-08-05
ci: move the npm release into GitHub Actions

venables:
## Changes

- Release workflow keeps a version-bump PR open on `main`
- Publish workflow ships to npm on manual dispatch
- Publish uses npm trusted publishing (OIDC), so the repo holds no npm token
- Publish waits for approval on a new `npm-publish` environment
- Pre-flight script lists package versions that npm does not have
- Tags survive a partial publish, and become annotated tag objects
- One GitHub release for each version, not for each package
- `bin/release` calls the correct clean script and stops on error
- `RELEASING.md` explains the flow, the setup, and the repair paths

## Problem

The release ran only from a laptop. `pnpm run publish:packages` needed an
interactive npm session, and account 2FA is set to `auth-and-writes`. The last
release stopped halfway. Three packages went to npm and five did not. A second
run fixed it, but nothing checked the tree before the publish, and nothing
recorded who approved the release.

`bin/release` also called `pnpm run Clean`. The script is `clean`, so the command
failed with `ERR_PNPM_NO_SCRIPT`. The file had no `set -e`, so the release
continued and skipped the clean step.

## Solution

The release runs in two phases, as changesets intends.

`release.yaml` runs on each push to `main`. It applies `changeset version` and
opens a version-bump PR on the `changeset-release/main` branch. The PR uses the
GitHub App identity that `audit-fix.yaml` already uses, so the check workflow
runs on it. The job stops early when no changesets are pending.

`publish.yaml` is the button. You dispatch it from `main` after the version PR
merges. The `verify` job refuses a dispatch from any other branch, lists the
package versions that npm does not have, refuses to continue when that list is
empty, and then runs the full check suite and `pnpm audit signatures`. The
`publish` job needs an approval on the `npm-publish` environment before it writes
to the registry.

### Authentication

npm trusted publishing. The repository stores no npm token. The publish job asks
GitHub for an OIDC id token, and pnpm exchanges it for a registry token that
lives for a few minutes.

pnpm does this itself and does not call the npm CLI. Its bundle carries
`libnpmpublish` and a first-party OIDC module at
`releasing/commands/lib/publish/oidc/`. `createPublishOptions` defaults to
`oidc = true`, so the exchange happens for every publish, and pnpm attaches
provenance on its own. This works through `changeset publish`, which runs
`pnpm publish` for each package. pnpm 11.7 already carries it, so the pinned
11.8.0 is enough.

No fallback token is configured, on purpose. pnpm treats a missing id token as a
soft failure: it logs `Skipped OIDC` and falls back to a configured token. A
token left in place would let a broken setup publish with it silently and
indefinitely. With no token the run fails instead, and a pre-flight check on
`ACTIONS_ID_TOKEN_REQUEST_URL` names the likely cause rather than leaving a bare
401.

### Recovery

A partial publish used to destroy the tags of the packages that did reach npm.
`changeset publish` tags the successes and only then exits non-zero, so a skipped
tag step lost them with the runner, and a retry could not recreate them because
`getUnpublishedPackages()` drops packages the registry already has. The tag step
now runs even when the publish fails.

The release version comes from the packages that actually ship, not from the
umbrella manifest. The `linked` group aligns the versions of packages in a
release, but it does not release every package every time — at 0.10.1, `caip`
stayed on 0.1.0 and `jwt`/`keys` on 0.9.0. Two guards refuse a tree that carries
more than one version, or that leaves the umbrella package out.

## Testing

- `actionlint` passes on all four workflow files.
- `pnpm run check:format` and `pnpm run knip` pass.
- `node .github/scripts/unpublished-packages.mjs` runs against the live registry
  and reports all eight packages at 0.11.0 as published.
- Both refusal guards were exercised directly: a mixed-version workspace and a
  workspace with no umbrella package each fail with the intended message.
- The `released` output was checked against a mixed fixture. It names the
  already-published package at the release version and omits the package on an
  older version.
- pnpm's OIDC support was confirmed by reading the 11.7.0 and 11.9.0 bundles:
  the token exchange endpoint, the first-party OIDC module, the `oidc = true`
  default, and the absence of any `run-npm` delegation.
- Five rounds of multi-agent review, to convergence.

The workflows cannot run end to end until the one-time setup is complete. See
`RELEASING.md`:

1. Create the `npm-publish` environment. Add required reviewers. Limit the
   deployment branches to `main`.
2. Add a trusted publisher to each of the eight packages on npmjs.com, naming
   repository `agentcommercekit/ack`, workflow `publish.yaml`, and environment
   `npm-publish`.

## AI usage

Claude Code (Opus 5) wrote the workflows, the pre-flight script, and the
documentation in this PR. I directed the design and reviewed each file.

## Summary by CodeRabbit

- **New Features**
  - Added automated workflows for versioning, package validation, approval-based publishing, tagging, and GitHub release creation.
  - Added detection and reporting for packages awaiting publication.
  - Added safeguards for partial, cancelled, or inconsistent releases.

- **Documentation**
  - Documented the Changesets-based release process, publishing requirements, fallback procedures, and contribution expectations.
  - Improved release command guidance and argument handling.

- **Bug Fixes**
  - Corrected the release cleanup command and enabled stricter error handling.


venables:
Closing for now

Issue #111, opened 2026-06-14
RFC: a scheme model for ACK-Pay (up-to, sessions/batch, subscriptions, streaming, refunds)


EfeDurmaz16:
@venables would love your read on the shape.
*Request for comment. Nothing is built against this yet. I want maintainer input on shape and naming before any code lands. Efe Baran Durmaz (@EfeDurmaz16).*

## TL;DR

ACK-Pay can express exactly one payment shape: a fixed amount, paid in full. x402 and MPP already ship the rest as first-class schemes and intents: spend caps, sessions, subscriptions, metering, refunds. This RFC proposes a small, additive `scheme` model for ACK-Pay plus five schemes ported from that prior art, each with a spec page, a reference implementation, and a conformance suite. ACK keeps its lead on verifiable receipts and identity, and gains the settlement shapes the rest of the ecosystem already speaks.

## The gap isn't trust. It's shape.

ACK already wins the two hard things its neighbors mostly lack: verifiable receipts (W3C VCs) and an agent-to-owner identity chain (ACK-ID). What it cannot say is "authorize up to X, settle less", "open a session and settle many micro-payments in one move", "charge on a cadence", "meter per unit", or "refund against a receipt".

None of those are exotic. x402 ships an explicit scheme registry (`exact`, `upto`, `deferred`, `escrow`). MPP ships an intent model (charge, session, subscription, vouchers). ACK-Pay ships one implicit shape: exact, paid in full.

An agent that can only pay a fixed amount in full is a demo, not an economy.

## What ACK-Pay expresses today

`PaymentOption` (`packages/ack-pay/src/schemas/valibot.ts:7-16`):

```ts
export const paymentOptionSchema = v.object({
  id: v.string(),
  amount: v.union([v.pipe(v.number(), v.integer(), v.gtValue(0)), v.string()]),
  decimals: v.pipe(v.number(), v.integer(), v.toMinValue(0)),
  currency: v.string(),
  recipient: v.string(),
  network: v.optional(v.string()),
  paymentService: v.optional(urlOrDidUri),
  receiptService: v.optional(urlOrDidUri),
})
```

There is no `scheme`, and no way to carry spend-cap, session, cadence, metering, or refund semantics. The receipt, though, already has a sanctioned extension point: `paymentReceiptClaimSchema.metadata` (`valibot.ts:34-37`). Each scheme can use it to record what was *actually* settled.

## Proposal: an additive `scheme` discriminator

[...]

PR #97, opened 2026-05-15
demo(payments): enforce policy before signing

EfeDurmaz16:
## Summary
- Add a tiny local payment policy guard for the payments demo
- Enforce policy after payment request verification but before returning an execution URL or signing the receipt-service payload
- Cover approved, approval-required, and denied policy outcomes with focused tests
- Document that this is an example pattern, not a normative ACK-Pay policy engine

Fixes #91

## Review follow-ups (addressed)
- Money handling now follows the repo-wide BigInt convention (`BigInt(paymentOption.amount)`), with an explicit positive-integer guard so fractional/non-positive/empty amounts the schema permits are denied instead of passing policy.
- The autonomous spend limit is now a per-currency map keyed on currency code and expressed in each currency's subunits, so the threshold is coherent across USD (2dp) / USDC (6dp); currencies with no configured limit are denied.
- `demoPaymentPolicy` is now the single source of truth — `enforcePaymentPolicy` spreads it and overrides only `allowedRecipients`, dropping the hardcoded literal.
- README reworded to state plainly that the amount check is an illustrative per-transaction cap, not a real spend control (split-gameable without a cumulative/rate budget).
- Optional cleanups: removed the dead `parsed` value and the unused `assertPaymentPolicyApproved` export.
- Rebased onto latest `main`.

## Verification
- pnpm run build
- pnpm --filter ./demos/payments check:types
- pnpm --filter ./demos/payments exec vitest run
- pnpm exec oxfmt --check demos/payments/README.md demos/payments/src/payment-policy.ts demos/payments/src/payment-policy.test.ts demos/payments/src/payment-service.ts
- pnpm exec oxlint demos/payments/src/payment-policy.ts demos/payments/src/payment-service.ts demos/payments/src/payment-policy.test.ts
- git diff --check

## AI Usage Disclosure
This contribution was AI-assisted. Initial work used Codex CLI and the Codex app; the review follow-ups above were completed with Claude Opus. AI assistance was used for repository/docs navigation, understanding ACK/Catena context, identifying relevant issues or contribution areas, and assisting with small edits/validation. I reviewed the final diff and take responsibility for the submitted changes.

## Summary by CodeRabbit

* **New Features**
  * Payments now enforce a policy-based approval flow that gates autonomous transactions, applies per-currency spending caps, and requires recipient approval before returning payment URLs or signing receipts.

* **Documentation**
[...]

PR #139, opened 2026-08-10
demo(payments): enforce a rolling spend budget, not just a per-payment cap

### LINKED ISSUE #138: demo(payments): add a cumulative spend budget to the policy guard
## Context

#97 added a policy-before-signing guard to the payments demo, and closed with an explicit caveat in both the code and the README:

> The amount check is an **illustrative per-transaction cap, not a real spend control.** A per-transaction limit is trivially defeated by splitting one payment into many smaller ones (`cap × N`). A production policy needs a cumulative and/or rate-limited budget (e.g. per-payer spend over a rolling window), not just a single-transaction threshold.

`demos/payments/src/payment-policy.ts` carries the same note. So the demo currently shows the right enforcement boundary (policy runs before execution or signing) with a check that does not actually bound spend.

## Proposal

Close that gap inside `demos/payments`, without touching the protocol or the published packages:

- A small in-memory ledger recording what a payer has put at risk inside a rolling window.
- An optional `budget: { windowMs, maxWindowAmount }` on `PaymentPolicy`, using the same per-currency subunit shape as the existing `maxAutonomousAmount`.
- An additive `authorizePayment` layer that runs the existing per-transaction checks first and then the window check, leaving `evaluatePaymentPolicy` unchanged.

Two details are what make it an enforcement mechanism rather than a display:

- **The check and the reservation have to be one synchronous step.** The Payment Service handler awaits `verifyPaymentRequestToken` before policy runs, so a guard that reads the running total and then writes to it lets two concurrent payments both observe the pre-payment total and both pass.
- **Reservations have to be keyed by payment attempt.** The Stripe path authorizes twice for a single payment — once for the payment URL, once on the callback — so a naive counter double-counts every payment. Keying on payment request id plus payment option id makes the second authorization re-check the window without consuming budget again, and gives a natural place to commit on receipt issuance or release when execution fails.

## Why this helps

- Makes the demo's spend control actually hold, instead of documenting that it does not.
- Keeps ACK's position unchanged: policy engines stay outside the protocol, and the demo shows where a real one plugs in.
- Gives a concrete, working consumer for the budget/approval shapes that #93 and #96 deferred for lack of one, and a reference point for the enforcement side of #111 without presuming any outcome for that RFC.

## Small first slice

- Everything under `demos/payments`; no package, schema, or protocol change and no new dependency.
- Unit tests for the split attack, window expiry, per-subject and per-currency isolation, idempotent re-authorization, and commit/release.
- README updated to describe the budget and to scope what is still demo-grade (in-memory storage, single-instance atomicity, denying rather than escalating to human approval).

Happy to adjust the shape — in particular whether a budget breach should return `denied` (mirroring the per-transaction cap today) or `approval_required`. I have this implemented and can open the PR if the direction looks right.

## AI Usage Disclosure

This issue was written with AI assistance (Claude Code, Opus), used for repository and history navigation and for drafting. I reviewed and understand the content and take responsibility for it.

mehmetkr-31:
## Summary

- Add a cumulative spend budget over a rolling window to the payments demo policy guard, backed by a small in-memory ledger in `src/spend-ledger.ts`
- Make the check-and-reserve a single synchronous step, so concurrent payments cannot both pass on the pre-payment total
- Key reservations by payment attempt, so the two calls of the Stripe flow authorize one payment once; commit on receipt issuance, release on failure
- Keep `evaluatePaymentPolicy` unchanged and put the new behaviour in an additive `authorizePayment` layer

Fixes #138.

This closes the gap #97 documented in its own code and README rather than adding anything new to the protocol surface:

> The amount check is an **illustrative per-transaction cap, not a real spend control.** A per-transaction limit is trivially defeated by splitting one payment into many smaller ones (`cap × N`). A production policy needs a cumulative and/or rate-limited budget (e.g. per-payer spend over a rolling window), not just a single-transaction threshold.

Everything stays inside `demos/payments`. No package, schema, or protocol change; no new dependency.

## What changed

**`src/spend-ledger.ts` (new)** — records what a subject has put at risk inside a rolling window, keyed by reference.

`reserve()` checks the window and writes in one synchronous step. This is the part that actually matters: the request handler awaits `verifyPaymentRequestToken` before policy runs, so a guard that read the running total and then wrote to it would let two concurrent payments both observe the pre-payment total and both pass. `commit()` marks a reservation settled; `release()` drops an unsettled one.

**`src/payment-policy.ts`** — `PaymentPolicy` gains an optional `budget: { windowMs, maxWindowAmount }`, using the same per-currency subunit shape as the existing `maxAutonomousAmount`. New `authorizePayment()` runs `evaluatePaymentPolicy` first and then the window check, so a payment rejected by the per-transaction cap or the allowlist never consumes budget. `evaluatePaymentPolicy` keeps its exact signature, behaviour, and reasons — the amount-parsing and per-currency lookup are extracted into two local helpers now shared by both paths.

**`src/payment-service.ts`** — both handlers authorize under `spendReference(paymentRequest.id, paymentOptionId)`, so the payment-URL call and the callback reserve once. The callback commits after the receipt is issued and releases if the Receipt Service call fails. The local `serverIdentity` in the callback is renamed to `payerIdentity`: it holds `PAYMENT_SERVICE_PRIVATE_KEY_HEX`, while `getTrustedRecipients` uses the same name for `SERVER_PRIVATE_KEY_HEX` two functions below.

**`README.md`** — replaces the "not a real spend control" warning with a "Rolling spend budget" section, and a new warning that scopes what is still demo-grade: in-memory storage, single-instance atomicity, denying rather than escalating to human approval, and tracking the demo's single autonomous payer because ACK-Pay carries no payer identity on the execution request.

## Notes on scope

- **Demo only, on purpose.** Per the discussion on #96, I did not add types or a budget model to `@agentcommercekit/ack-pay`. If a shape like this is ever worth publishing, this is the concrete consumer that would justify it.
- **Related to but not implementing #111.** That RFC proposes a protocol-level `scheme` model (`upto`, sessions, subscriptions). This is a Payment Service enforcing its own budget, which is a different layer; nothing here presumes an outcome for #111.
- **`denied` rather than `approval_required`** for a budget breach, mirroring how the per-transaction cap already behaves. The README notes that a real service would escalate instead.

## Verification

```
pnpm run build
pnpm run check                                     # 29/29 tasks, 0 lint warnings/errors
pnpm --filter ./demos/payments exec vitest run     # 34 passed (19 policy, 15 ledger)
pnpm run knip                                      # clean
git diff --check
```

Tests cover the split attack (three below-cap payments approved, the fourth denied), window expiry, per-subject and per-currency isolation, idempotent re-authorization, reservations not being consumed by denied or approval-required payments, commit/release semantics, and the no-budget-configured path.

I also ran the real handlers over HTTP with the Receipt Service live, driving the Stripe path end to end:

| | result |
|---|---|
| happy path at the demo's 5-subunit amount | `initiate=200 callback=200`, receipt issued |
| three payments at the per-transaction cap | `200, 200, 200` |
| fourth payment, crossing the window budget | `403 Payment exceeds the autonomous spend budget for the current window` |
| re-authorizing an already-reserved attempt | `200`, no additional budget consumed |
| the denied payment after that replay | still `403` |

## AI Usage Disclosure

This contribution was AI-assisted using Claude Code (Opus). AI assistance was used for repository and history navigation, reviewing the existing policy guard and its review discussion, drafting the ledger and tests, and running verification. I reviewed the final diff, understand how the reserve/commit/release path interacts with the two-phase Stripe flow and why the check and reserve must not be separated by an await, and take responsibility for the submitted changes.

## Summary by CodeRabbit

* **New Features**
  * Added rolling spend budgets for payment subjects, with configurable currency-specific and per-transaction limits.
  * Payment authorization now supports approval requirements for unknown recipients and prevents duplicate budget consumption.
  * Payment receipts are securely tied to payer identity, with reserved funds released when receipt retrieval fails.
  * Added support for committing successful payment reservations and tracking spending within a time window.

* **Documentation**
  * Updated the payment demo policy documentation with budget and reservation behavior.

### DIFF
--- a/demos/payments/README.md
+++ b/demos/payments/README.md
@@ -20,7 +20,7 @@ This interactive command-line demo showcases a common use case: the **Server-Ini
   - Handling currency conversions.
   - Integrating compliance checks (KYC/AML).
   - Facilitating complex payment routing.
-  - Enforcing local payment policy before returning an execution URL or signing a receipt-service payload.
+  - Enforcing local payment policy, including a cumulative spend budget, before returning an execution URL or signing a receipt-service payload.
 
 You can learn more about the full ACK-Pay protocol at [www.agentcommercekit.com](https://www.agentcommercekit.com).
 
@@ -36,9 +36,10 @@ Payment Services should replace it with their own owner, risk, compliance, or
 human-approval system. The important safety boundary is that policy enforcement
 happens before execution or signing:
 
-- known low-value recipient: continue automatically
+- known low-value recipient, within budget: continue automatically
 - unknown recipient: return `approval_required`
-- amount above the illustrative per-transaction cap: deny before payment
+- amount above the per-transaction cap: deny before payment execution
+- amount that would exceed the rolling spend budget: deny before payment
   execution
 
 The per-currency cap is expressed in each currency's smallest subunit, so a
@@ -46,12 +47,41 @@ single flat threshold is never compared across currencies with different
 decimals (e.g. USD at 2dp vs USDC at 6dp). Currencies without a configured
 limit are denied outright.
 
+### Rolling spend budget
+
+A per-transaction cap on its own is not a spend control: it is trivially
+defeated by splitting one payment into many smaller ones (`cap × N`). The demo
+policy therefore also carries a cumulative budget over a rolling window,
+enforced against the small in-memory ledger in `src/spend-ledger.ts`.
+
+- `maxAutonomousAmount` bounds a single payment.
+- `budget.maxWindowAmount` bounds their sum over `budget.windowMs`, in the same
+  per-currency subunits.
+
+Two details are what make the budget hold rather than merely look right:
+
+- **The check and the reservation are one synchronous step.** The request
+  handler awaits token verification before policy runs, so a guard that read the
+  running total and then wrote to it would let two concurrent payments both
+  observe the pre-payment total and both pass. `SpendLedger.reserve` does both
+  at once.
+- **Reservations are keyed by payment attempt.** The Stripe path authorizes
+  twice for a single payment — once for the payment URL, once on the callback —
+  so reservations are keyed by payment request id plus payment option id, and
+  the second authorization re-checks the window without counting the payment
+  twice. A reservation is committed once the receipt is issued, and released if
+  the Receipt Service call fails.
+
 > [!IMPORTANT]
-> The amount check is an **illustrative per-transaction cap, not a real spend
-> control.** A per-transaction limit is trivially defeated by splitting one
-> payment into many smaller ones (`cap × N`). A production policy needs a
-> cumulative and/or rate-limited budget (e.g. per-payer spend over a rolling
-> window), not just a single-transaction threshold.
+> This is still a demo, not production spend control. The ledger lives in
+> process memory, so it resets with the demo and is not shared between Payment
+> Service instances; a real budget needs durable storage and an atomic
+> check-and-reserve (a transactional `UPDATE ... WHERE`, or a distributed lock)
+> across every instance that can authorize payments. A real service would also
+> route a budget breach to human approval rather than denying outright, and
+> would key the budget on its authenticated payer — ACK-Pay carries no payer
+> identity on the payment execution request, so the demo tracks the single
+> autonomous payer it spends as.
 
 The demo allowlist is based on the configured server identity, not the issuer
 claimed by each incoming Payment Request token. A real Payment Service should
--- a/demos/payments/README.md
+++ b/demos/payments/README.md
@@ -65,12 +65,14 @@ Two details are what make the budget hold rather than merely look right:
   running total and then wrote to it would let two concurrent payments both
   observe the pre-payment total and both pass. `SpendLedger.reserve` does both
   at once.
-- **Reservations are keyed by payment attempt.** The Stripe path authorizes
+- **Reservations are keyed by payment execution.** The Stripe path authorizes
   twice for a single payment — once for the payment URL, once on the callback —
-  so reservations are keyed by payment request id plus payment option id, and
-  the second authorization re-checks the window without counting the payment
-  twice. A reservation is committed once the receipt is issued, and released if
-  the Receipt Service call fails.
+  so the Payment Service mints an execution id at initiation and carries it
+  through the callback URL. The second authorization then re-checks the window
+  without counting the payment twice, while a second execution of the same
+  Payment Request is charged separately, since it moves money separately. A
+  reservation is committed once the receipt is issued, and released if signing
+  or the Receipt Service call fails.
 
 > [!IMPORTANT]
 > This is still a demo, not production spend control. The ledger lives in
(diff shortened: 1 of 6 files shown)


mehmetkr-31:
Thanks both — all three findings were real. Pushed `5414824` and `6100986`; the reviews above were against `a6ce66e`.

### Reserve per payment execution, not per Payment Request (@chatgpt-codex-connector)

This was the important one, and it undercut the point of the PR. Keying on the Payment Request meant a request presented for execution twice consumed the budget once, so the agent the budget is meant to bound could bypass it by re-executing one request.

The Payment Service now mints an execution id at initiation and carries it through the callback URL, so the two calls of one execution still reserve once while a second execution is charged separately. The demo client needs no change — it forwards whatever `return_to` it is given — and `return_to` is now `encodeURIComponent`'d since it carries a query of its own. A callback that arrives with an unknown execution id reserves again rather than escaping the budget, so the unhappy path fails closed; a real Payment Service would resolve the execution from the provider's session id server-side instead of trusting the caller to echo it back, which is noted at the call site.

### Release the reservation when JWT creation fails (@coderabbitai)

Correct — `createJwt` sat outside the try, so a signing failure left a reservation with nothing to commit or release. It is inside now.

### Document the post-capture denial path (@coderabbitai)

Also correct, and documentation rather than a code change is the right call here: the callback re-authorizes after the card is captured, so a denial there stops the receipt without stopping the charge. Added to the README limitations block, including that a real service should settle and flag an over-budget capture for reconciliation rather than refuse it.

### Also fixed, self-reported

The reservation key was `${paymentRequestId}:${paymentOptionId}`. Both parts arrive as unconstrained strings on the Payment Request, so `("a:b", "c")` and `("a", "b:c")` collided onto one entry and the second amount fell out of the window — reachable by any payment request that clears the recipient allowlist. The key is encoded now, and `spendReference` moved next to the ledger so it is testable without importing the Payment Service, which starts a server on import.

### Verification

```
pnpm run build
pnpm run check                                     # 29/29 tasks, 0 lint warnings/errors
pnpm --filter ./demos/payments exec vitest run     # 40 passed (19 policy, 21 ledger)
pnpm run knip                                      # clean
git diff --check
```

Re-ran the three services and drove the Stripe path over HTTP, extracting `return_to` from the payment URL exactly as `demos/payments/src/index.ts` does:

| | result |
|---|---|
| happy path at the demo's 5-subunit amount | `initiate=200`, execution id present in `return_to`, `callback=200`, receipt issued |
| replaying the callback for one execution | `200`, no additional budget consumed |
| three payments at the per-transaction cap | `200, 200, 200` |
| re-executing an already-executed request | `403` — a `200` here would mean re-execution is free, which is the bug above |
| a fresh payment on the exhausted budget | `403` |

### AI Usage Disclosure

These follow-ups were prepared with AI assistance (Claude Code, Opus), used for navigating the review threads, drafting the changes and tests, and running verification. I reviewed the final diff and take responsibility for it.

venables:
Thanks for the work here @mehmetkr-31 , this is clearly a lot of effort. Closing it on scope: the per-payment cap in the demo is deliberately illustrative (see the README note that came out of #97), and the demo's job is to show where enforcement happens, not to be a spend-control product. A two-phase reservation ledger plus 400 lines of tests is more surface than we want to maintain in a demo.

If you think a budget example is worth having, ideally we could do something in a ~100 line version first (one map, one window check).

PR #140, opened 2026-08-10
test(vc): add tests for the isVerifiable proof guard

devorun:
## Summary

`isVerifiable` (`packages/vc/src/verification/is-verifiable.ts`) was the only helper on the verification path in `@agentcommercekit/vc` without a co-located test. It guards both `verify-parsed-credential` and `is-revoked`, so pinning down its proof shape contract is worthwhile.

This adds a unit test suite that covers each branch of the shape check:

- returns `true` for a proof object carrying a `type` (with and without extra fields)
- returns `false` when the proof is missing, `null`, `undefined`, or not an object
- returns `false` when the proof is an empty object or has no `type`

Tests only — no source changes.

## Testing

- `pnpm --filter @agentcommercekit/vc test` — the vc suite passes, including the 8 new cases
- `oxlint` and `oxfmt --check` are clean for the new file

## AI usage disclosure

Per `AI_POLICY.md`: these tests were written with AI assistance (Claude Code). I reviewed every case and understand the behavior each one verifies.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for credential verifiability checks.
  * Verified handling of valid typed proofs and invalid, missing, empty, or improperly formatted proofs.

### DIFF
--- a/packages/vc/src/verification/is-verifiable.test.ts
+++ b/packages/vc/src/verification/is-verifiable.test.ts
@@ -1,0 +1,60 @@
+import { describe, expect, it } from "vitest"
+
+import type { W3CCredential } from "../types"
+import { isVerifiable } from "./is-verifiable"
+
+// A minimal valid W3C credential, without a proof
+const baseCredential: W3CCredential = {
+  "@context": ["https://www.w3.org/2018/credentials/v1"],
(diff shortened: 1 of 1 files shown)


venables (approved):
Thanks @devorun

PR #141, opened 2026-08-11
test(vc): add tests for the credential verification errors

devorun:
## Summary

`packages/vc/src/verification/errors.ts` defines the public `CredentialVerificationError` hierarchy but had no direct tests. These errors carry real contracts that the rest of the verification code relies on, so this pins them down:

- every error extends `CredentialVerificationError` (which API error handlers catch) and `Error`
- each error exposes its documented default message and accepts a custom one
- `RevocationCheckError` and `UnsupportedCredentialStatusError` store their `detail` for logging but keep it **out** of `message` — the detail names a dereferenced URL or a credential-supplied value, and handlers return `message` to the caller
- `RevocationCheckError` propagates its `cause`

Tests only — no source changes.

## Testing

- `pnpm --filter @agentcommercekit/vc test` — the vc suite passes, including the new cases
- `oxlint` and `oxfmt --check` are clean for the new file

## AI usage disclosure

Per `AI_POLICY.md`: these tests were written with AI assistance (Claude Code). I reviewed every case and understand the behavior each one verifies.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for credential verification errors.
  * Verified default and custom messages, error names, inheritance, optional details, and cause propagation.

### DIFF
--- a/packages/vc/src/verification/errors.test.ts
+++ b/packages/vc/src/verification/errors.test.ts
@@ -1,0 +1,169 @@
+import { describe, expect, it } from "vitest"
+
+import {
+  CredentialExpiredError,
+  CredentialRevokedError,
+  CredentialVerificationError,
+  InvalidControllerClaimError,
+  InvalidCredentialError,
+  InvalidCredentialSubjectError,
+  InvalidProofError,
+  RevocationCheckError,
+  UnsupportedCredentialStatusError,
+  UnsupportedCredentialTypeError,
+  UnsupportedProofTypeError,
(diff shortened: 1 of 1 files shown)


venables:
Thanks for the contribution. Closing this one: the error classes here are one-line subclasses, and the behavior that matters (detail staying out of message) is already asserted across `is-revoked.test.ts` and the verify tests. Pinning 40+ default-message strings means every future copy edit breaks tests without catching a real regression.

PR #143, opened 2026-08-12
fix(caip): validate full string in caip2Parts and caip10Parts

bayrakdarerdem:
## Summary

`caip2Parts` and `caip10Parts` parsed input with `caip.split(":")` and only
checked that the destructured segments were non-empty. `split()` doesn't
cap the number of segments, so malformed input with extra trailing
`:segments` (e.g. `"eip155:1:extra"`) silently parsed as if truncated,
instead of being rejected.

`caip10Parts` is reached from `@agentcommercekit/did`'s `did:pkh`
`createVerificationMethod`, via the public
`createDidPkhDocumentFromCaip10AccountId`. That entry point doesn't itself
re-validate its input against the CAIP-10 pattern, so a malformed account
ID passed there could previously produce a DID document built from
silently truncated account data instead of throwing.

## Fix

Both functions now test the input against their full regex
(`caip2ChainIdRegex` / `caip10AccountIdRegex`) before parsing, so
malformed input is rejected up front instead of silently truncated.

## Tests

- Added regression tests for both functions (`caip10Parts` previously had
  no test coverage at all).
- Ran the full `caip` and `did` package test suites — all passing.
- Ran lint/format checks — clean.

## Changeset

Included (`@agentcommercekit/caip` patch).

---
Assisted by Claude (Anthropic) — the bug and fix were found and
implemented working interactively through the codebase and tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for CAIP-2 chain IDs and CAIP-10 account IDs.
  * Malformed identifiers with extra segments or missing account addresses are now rejected instead of being partially parsed.
  * Prevented invalid CAIP-10 identifiers from producing incorrect verification methods.

* **Tests**
  * Added coverage for valid EVM and Solana account IDs and malformed chain/account identifiers.

### DIFF
--- a/packages/caip/src/caips/caip-10.ts
+++ b/packages/caip/src/caips/caip-10.ts
@@ -48,6 +48,15 @@ export function createCaip10AccountId(
 }
 
 export function caip10Parts(caip: Caip10AccountId): Caip10AccountIdParts {
+  // Validate the full string against the CAIP-10 pattern rather than just
+  // splitting on ":". A naive split silently drops any segments after the
+  // third colon (e.g. "eip155:1:0xabc:extra" would parse as if it were
+  // just "eip155:1:0xabc"), so malformed input could pass through
+  // undetected instead of being rejected.
+  if (!caip10AccountIdRegex.test(caip)) {
+    throw new Error("Invalid CAIP-10 account ID")
+  }
+
   const [namespace, reference, accountId] = caip.split(":")
   if (!namespace || !reference || !accountId) {
     throw new Error("Invalid CAIP-10 account ID")
--- a/packages/caip/src/caips/caip-2.ts
+++ b/packages/caip/src/caips/caip-2.ts
@@ -38,6 +38,15 @@ export const caip2ChainIds = {
 } as const
 
 export function caip2Parts(caip: Caip2ChainId): Caip2ChainIdParts {
+  // Validate the full string against the CAIP-2 pattern rather than just
+  // splitting on ":". A naive split silently drops any segments after the
+  // second colon (e.g. "eip155:1:extra" would parse as if it were just
+  // "eip155:1"), so malformed input could pass through undetected instead
+  // of being rejected.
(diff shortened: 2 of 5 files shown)


packages/caip/src/caips/caip-10.ts
```diff
@@ -48,6 +48,15 @@ export function createCaip10AccountId(
 }
 
 export function caip10Parts(caip: Caip10AccountId): Caip10AccountIdParts {
+  // Validate the full string against the CAIP-10 pattern rather than just
+  // splitting on ":". A naive split silently drops any segments after the
+  // third colon (e.g. "eip155:1:0xabc:extra" would parse as if it were
+  // just "eip155:1:0xabc"), so malformed input could pass through
+  // undetected instead of being rejected.
```
venables, line 55: we should drop these inline comments

packages/caip/src/caips/caip-2.ts
```diff
@@ -38,6 +38,15 @@ export const caip2ChainIds = {
 } as const
 
 export function caip2Parts(caip: Caip2ChainId): Caip2ChainIdParts {
+  // Validate the full string against the CAIP-2 pattern rather than just
+  // splitting on ":". A naive split silently drops any segments after the
+  // second colon (e.g. "eip155:1:extra" would parse as if it were just
+  // "eip155:1"), so malformed input could pass through undetected instead
+  // of being rejected.
```
venables, line 45: please drop these inline comments

packages/caip/src/caips/caip-2.ts
venables, line 54: This check and throw are now effectively dead since the anchored regex guarantees exactly two or three non-empty segments.

packages/caip/src/caips/caip-10.ts
venables, line 63: This check and throw are now effectively dead since the anchored regex guarantees exactly two or three non-empty segments.

venables (changes requested):
thanks! just a couple comments for maintainability

PR #149, opened 2026-08-13
fix: validate decimals and fail closed on unparseable expirationDate

### LINKED ISSUE #147: bug: `paymentOptionSchema.decimals` uses `toMinValue` (clamps) instead of `minValue` (validates)
## Summary

## Affected file

## Steps to reproduce

```ts
import * as v from "valibot"

  id: "opt-1",
  amount: 100,
  currency: "USDC",
})

```

## Root cause

```ts
```

## Proposed fix

```ts
```

### LINKED ISSUE #148: bug: `isExpired` returns `false` for unparseable `expirationDate` (fail open)
## Summary

## Affected file

## Steps to reproduce

```ts

const credential = {
  credentialSubject: {},
}

```

## Root cause

```ts
// is-expired.ts
  return false   // ← fail-open
}
```

```ts
}
```

## Impact

## Proposed fix

```ts
}
```

Sertug17:
Closes #147
Closes #148

### `packages/ack-pay/src/schemas/valibot.ts`
Replace `v.toMinValue(0)` with `v.minValue(0)` for the `decimals` field.
`toMinValue` is a *transformation* that silently clamps negative values to 0;
`minValue` is a *validation* that rejects them with an error. A `decimals`
value of -3 should never silently become 0.

### `packages/vc/src/verification/is-expired.ts`
Return `true` (expired) instead of `false` when `expirationDate` is present
but unparseable. Returning `false` allows a credential with a garbage date to
pass all downstream verification checks the wrong default for a
security-critical guard. Failing closed is safer here.

---

> AI disclosure (per AI_POLICY.md): This fix was developed with Claude AI assistance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Payment option decimal values are now correctly rejected when below zero.
  * Credentials with invalid or empty expiration dates are now treated as expired, improving verification accuracy.

### DIFF
--- a/packages/ack-pay/src/payment-request.test.ts
+++ b/packages/ack-pay/src/payment-request.test.ts
@@ -40,4 +40,17 @@ describe("isPaymentRequest", () => {
   it("returns false if given a non-object", () => {
     expect(isPaymentRequest(1)).toBe(false)
   })
+
+  it("returns false when decimals is negative", () => {
+    // decimals uses minValue(0) — a validation that rejects negative values.
+    // The previous toMinValue(0) silently clamped -6 to 0 instead of rejecting it.
+    expect(
+      isPaymentRequest({
+        ...validPaymentRequest,
+        paymentOptions: [
+          { ...validPaymentRequest.paymentOptions[0], decimals: -6 },
+        ],
+      }),
+    ).toBe(false)
+  })
 })
--- a/packages/ack-pay/src/schemas/valibot.ts
+++ b/packages/ack-pay/src/schemas/valibot.ts
@@ -7,7 +7,7 @@ const urlOrDidUri = v.union([v.pipe(v.string(), v.url()), didUriSchema])
 export const paymentOptionSchema = v.object({
   id: v.string(),
   amount: v.union([v.pipe(v.number(), v.integer(), v.gtValue(0)), v.string()]),
-  decimals: v.pipe(v.number(), v.integer(), v.toMinValue(0)),
+  decimals: v.pipe(v.number(), v.integer(), v.minValue(0)),
   currency: v.string(),
   recipient: v.string(),
   network: v.optional(v.string()),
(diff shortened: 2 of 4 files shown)


qlxjcj (commented):
Both fixes are in the right direction. I verified them locally:

- `packages/ack-pay/src/schemas/valibot.ts` — swapping `v.toMinValue(0)` for `v.minValue(0)` correctly turns a silent clamp into a validation, matching the Zod side's `.nonnegative()`. Good.
- `packages/vc/src/verification/is-expired.ts` — returning `true` for an unparseable `expirationDate` makes `isExpired` fail closed, consistent with `is-revoked.ts`. Good.

Two things to address before this merges:

**1. Unrelated README changes (6 files) belong in a separate PR.**
This PR bundles a "docs: fix missing Zod v4 import examples" commit (from May 20) together with the two bug fixes. That commit rewrites import-order examples across `ack-id`, `ack-pay`, `caip`, `did`, `jwt`, and `vc` READMEs. It's a legitimate doc fix, but it's unrelated to #147/#148, adds review surface, and the commit message doesn't mention the bugs. Please split it out.

**2. Missing regression coverage for the fixes.**
- The valibot change has no test asserting a negative `decimals` is rejected. `packages/ack-pay/src/payment-request.test.ts` covers `isPaymentRequest` shape but never exercises `decimals: -6`. Without it, the `toMinValue` vs `minValue` distinction can silently regress.
- The `isExpired` fix doesn't handle the empty-string variant. `if (!credential.expirationDate)` treats `expirationDate: ""` as absent and returns `false` (fail-open), since `!""` is `true`. The empty string is present but unparseable (`new Date("")` -> `NaN`), so it never reaches the new fail-closed branch. Consider:

```diff
-  if (!credential.expirationDate) {
+  if (credential.expirationDate === undefined) {
     return false
   }
```

with a regression test for the empty string. (CodeRabbit raised the same point on #154.)

**3. Branch is behind `main`.**
The `is-expired.test.ts` hunk's base (`5a67bf7`) differs from current `main` (`ac456e4`), and the PR shows as `DIRTY`. Please rebase.

Also note: #154 is a separate PR fixing just #148. If you want to avoid three PRs touching the same `isExpired` function, coordinating with it (or closing one) would help maintainers.

PR #67, opened 2026-03-25
fix(caip): add input validation to createCaip10AccountId

ak68a:
## Summary
- Validate `chainId` against `caip2ChainIdRegex` and `address` against `caip10AccountAddressRegex` before creating the account ID
- Add test cases for invalid inputs

`createCaip10AccountId` previously performed no validation, allowing malformed chain IDs and addresses to propagate silently. The regex patterns were already defined in the module but not being used. This brings the function in line with `caip10Parts()`, which already validates its input.

## Test plan
- [x] Existing CAIP tests pass
- [x] New tests verify rejection of invalid chain IDs and addresses

---

**AI Disclosure:** This PR was developed with assistance from Claude Code (Claude Opus).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when creating CAIP-10 account IDs so invalid chain identifiers and empty/invalid account addresses are rejected with clearer error responses.
* **Tests**
  * Added test coverage for the new validation and expected failure cases to ensure reliable behavior.

PR #150, opened 2026-08-13
fix(caip): reject CAIP-2/CAIP-10 IDs with extra colon-delimited segments

batuhankocyigit:
## Bug

`caip2Parts` and `caip10Parts` parse their input with `caip.split(":")` and
destructure the expected number of parts, but never check whether there
were *more* parts than expected. A string like `"eip155:1:evil"` is invalid
per the CAIP-2 spec (the reference component `[-_a-zA-Z0-9]{1,32}` cannot
contain a colon), but `caip2Parts` currently accepts it, silently dropping
the "evil" segment and returning `{ namespace: "eip155", reference: "1" }`
instead of throwing. Same issue in `caip10Parts` with a 4th segment.

This mirrors the exact bug class fixed in `createCaip10AccountId` in #67 —
that fix validates with `caip2ChainIdRegex`/`caip10AccountAddressRegex`
before constructing an ID, but the "Parts" parsing functions going the
other direction (string → components) were never covered, and have no
tests for malformed multi-segment input.

Both functions are exported from `@agentcommercekit/caip`'s public API, so
downstream consumers calling them directly on untrusted input (e.g. a CAIP
string embedded in a DID or credential) would get a silently-truncated
parse instead of a validation error.

Current call sites in this repo (`packages/did/src/methods/did-pkh.ts`)
happen to already validate upstream with `isCaip10AccountId` /
`createCaip10AccountId`, so this isn't exploitable through them today —
this is a defense-in-depth / spec-conformance fix for the exported function
itself.

## Fix

Both functions now also check that `split(":")` produced exactly the
expected number of segments, throwing the existing error message otherwise.

## Testing

Added tests reproducing the bug (confirmed they fail on unpatched `main`:
"expected [Function] to throw an error"), passing with the fix. Also added
missing basic coverage for `caip10Parts`, which had none before.

- `packages/caip`: 86/86 passing (was 27/27 before this PR's added tests)
- `packages/did` (the package that actually calls `caip10Parts`): 74/74
  passing, no regressions
- `oxlint` and `tsc --noEmit`: clean

## AI usage disclosure

Per AI_POLICY.md: this fix was found and implemented with Claude (Anthropic).
I asked it to review the caip package for validation gaps given the recent
#67 fix in the same area; it identified the split-based truncation issue,
wrote the failing tests, confirmed them against unpatched `main`, then
applied and verified the fix and ran the affected package + downstream
consumer test suites. I've reviewed the diff and understand the bug and
the fix — happy to answer any questions about it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for chain and account identifiers.
  * Invalid identifiers containing extra colon-separated segments are now rejected.
  * Existing validation for missing namespaces or references remains intact.

* **Tests**
  * Added coverage for valid, incomplete, empty, and malformed identifier inputs.

### DIFF
--- a/packages/caip/src/caips/caip-10.test.ts
+++ b/packages/caip/src/caips/caip-10.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest"
 
-import { createCaip10AccountId } from "./index"
+import { caip10Parts, createCaip10AccountId } from "./index"
 
 describe("createCaip10AccountId", () => {
   it("creates a caip 10 account ID for EVM address", () => {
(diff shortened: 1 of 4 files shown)


venables:
Thanks for the contribution. Closing this as a duplicate of #143.

PR #154, opened 2026-08-13
fix(vc): isExpired fails closed on an unparseable expirationDate

### LINKED ISSUE #148: bug: `isExpired` returns `false` for unparseable `expirationDate` (fail open)

ygd58:
Fixes #148

`isExpired` returned `false` (not expired) whenever `credential.expirationDate` was present but could not be parsed into a valid date. Since `isExpired` is the check `verifyParsedCredential` uses to reject expired credentials, a credential with a malformed or malicious `expirationDate` value was treated as never-expiring instead of being rejected.

**Fix:** `isExpired` now fails closed — an unparseable `expirationDate` is treated as expired, matching the safer default for a security-relevant check. Updated the existing test that asserted the old fail-open behavior.

**Changeset:** added (`@agentcommercekit/vc`, patch).

Verified locally: `pnpm --filter @agentcommercekit/vc test -- is-expired` (5/5 passing), `oxlint` and `oxfmt --check` clean on both changed files.

---

**AI usage disclosure** (per AI_POLICY.md): this fix was developed with Claude (Anthropic) assistance — identifying the bug, writing the fix, updating the test, and verifying locally. I reviewed and understand the change: it flips a single boolean return value in one function so a credential with an unparseable expiration date is rejected instead of silently accepted, and updates the one test that covered that branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Credential expiration checks now fail closed: malformed, empty, or non-string expiration values are treated as expired.
  * Credentials with invalid expiration values can no longer bypass expiration validation.

* **Documentation**
  * Clarified how expiration checks handle missing, malformed, empty, and non-string date values.
  * Clarified expiration handling for status list credentials, including expiration at the current time.

### DIFF
--- a/packages/vc/src/verification/is-expired.ts
+++ b/packages/vc/src/verification/is-expired.ts
@@ -3,8 +3,13 @@ import type { W3CCredential } from "../types"
 /**
  * Check if a credential is expired
  *
+ * Fails closed: a credential with an `expirationDate` that is present but
+ * cannot be parsed as a valid date is treated as expired, not as
+ * non-expiring.
+ *
  * @param credential - The {@link W3CCredential} to check
- * @returns `true` if the credential is expired, `false` otherwise
+ * @returns `true` if the credential is expired (or has an unparseable
+ *   expiration date), `false` otherwise
  */
 export function isExpired(credential: W3CCredential): boolean {
   if (!credential.expirationDate) {
@@ -14,8 +19,11 @@ export function isExpired(credential: W3CCredential): boolean {
   const expirationDate = new Date(credential.expirationDate)
 
   if (isNaN(expirationDate.getTime())) {
-    // Expiration date is invalid, so we consider the credential not expired
-    return false
+    // Expiration date is present but unparseable. Fail closed: an
+    // unparseable expiration date must not be treated as "never expires",
+    // since that would let a malformed or malicious `expirationDate` value
+    // grant a credential unbounded validity.
+    return true
   }
 
   return expirationDate < new Date()
--- a/packages/vc/src/verification/is-expired.ts
+++ b/packages/vc/src/verification/is-expired.ts
@@ -16,6 +16,16 @@ export function isExpired(credential: W3CCredential): boolean {
     return false
   }
 
+  // `parseJwtCredential` does not validate that `expirationDate` is a
+  // string, so a malformed or malicious credential could carry a number
+  // (which `Date()` accepts as epoch milliseconds) or another non-string
+  // JSON value here. Reject anything that isn't a string outright, rather
+  // than letting it reach `new Date()`, which would silently accept types
+  // the {@link W3CCredential} type only documents as a string.
+  if (typeof credential.expirationDate !== "string") {
+    return true
+  }
+
   const expirationDate = new Date(credential.expirationDate)
 
   if (isNaN(expirationDate.getTime())) {
--- a/packages/vc/src/verification/is-expired.ts
+++ b/packages/vc/src/verification/is-expired.ts
@@ -12,7 +12,7 @@ import type { W3CCredential } from "../types"
  *   expiration date), `false` otherwise
  */
 export function isExpired(credential: W3CCredential): boolean {
-  if (!credential.expirationDate) {
+  if (credential.expirationDate === undefined) {
     return false
   }
 
--- a/.changeset/is-expired-fail-closed.md
+++ b/.changeset/is-expired-fail-closed.md
@@ -1,0 +1,14 @@
+---
+"@agentcommercekit/vc": patch
+---
+
+Fix `isExpired` failing open on an unparseable `expirationDate`
+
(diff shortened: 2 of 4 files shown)


qlxjcj (commented):
Verified the fix: flipping the unparseable branch to `return true` makes `isExpired` fail closed, and the updated test (`treats an unparseable expiration date as expired (fail closed)`) matches. I also ran the change locally against @agentcommercekit/vc — all 109 tests pass.

CodeRabbit's open comment about the empty string is valid and worth fixing here, since this PR already touches the same guard. `if (!credential.expirationDate)` treats `expirationDate: ""` as absent and returns `false` (fail-open), because `!""` is `true`. The empty string is present but unparseable (`new Date("")` -> `NaN`), so it never reaches the new fail-closed branch.

Suggested change:

```diff
-  if (!credential.expirationDate) {
+  if (credential.expirationDate === undefined) {
     return false
   }
```

with a regression test asserting an empty `expirationDate` is treated as expired. Otherwise the fix leaves the same malformed-input hole open for the empty-string variant.

ygd58:
Good catch, thank you - pushed a fix. Changed the guard from `!credential.expirationDate` to `credential.expirationDate === undefined`, so an empty string reaches the `isNaN` fail-closed branch instead of short-circuiting on the falsy check. Added the regression test you suggested (`empty-string expirationDate is treated as expired`). Verified locally: 6/6 tests passing, oxlint and oxfmt clean.

packages/vc/src/verification/is-expired.ts
```diff
@@ -3,19 +3,27 @@ import type { W3CCredential } from "../types"
...
+ * non-expiring.
+ *
  * @param credential - The {@link W3CCredential} to check
- * @returns `true` if the credential is expired, `false` otherwise
+ * @returns `true` if the credential is expired (or has an unparseable
+ *   expiration date), `false` otherwise
  */
 export function isExpired(credential: W3CCredential): boolean {
-  if (!credential.expirationDate) {
+  if (credential.expirationDate === undefined) {
```
venables, line 15: The guard only excludes `undefined`, so a non-string JSON value (a number, a one-element array) forms a valid `Date` and returns "not expired"; `parseJwtCredential` does not validate `expirationDate`, so such values can reach here. Verified with node. 

The security gain is small since a valid far-future string does the same. 

Potential fix: return `true` when `typeof credential.expirationDate !== "string"`, plus one test with a numeric value.

venables (changes requested):
This change has some downstream effects worht fixing:

- [packages/vc/src/verification/is-revoked.ts:414-416](https://github.com/agentcommercekit/ack/blob/ab564b2d432c462477f08b14841e2abf6c48e816/packages/vc/src/verification/is-revoked.ts#L414-L416) — the comment says the direct check exists because `isExpired` "reads an unparseable date as 'not expired'"; this PR makes that statement false. We should reword to the reasons that still hold (ex: this path must throw `undetermined` with a URL-specific message, and must reject a list that expires at exactly `now`).

- [packages/vc/src/verification/verify-parsed-credential.ts:73-75](https://github.com/agentcommercekit/ack/blob/ab564b2d432c462477f08b14841e2abf6c48e816/packages/vc/src/verification/verify-parsed-credential.ts#L73-L75) — an unreadable `expirationDate` now surfaces as the generic "Credential is expired" message, while the sibling check in `is-revoked.ts:421-423` says "unreadable expirationDate". Might be worth being consistent pass a distinct message, or leave as-is for the smaller diff.

ygd58:
Thanks for the thorough review - all three addressed:

1. **Non-string `expirationDate` (the real bug)**: fixed exactly as suggested - added `typeof credential.expirationDate !== "string"` returning `true`. Added a regression test using a numeric epoch-ms value *ten years in the future*, deliberately not a small number like `1` - a small value would coincidentally "pass" via the normal date-comparison path regardless of whether the guard exists (since `new Date(1)` is already far in the past), which would give false confidence. A future-dated epoch number only fails without the explicit type check, so it actually exercises the fix.

2. **Stale comment in `is-revoked.ts`**: reworded to the reasons that direct check still needs to exist now - it must throw `undetermined()` with a URL-specific message (`isExpired` only returns a boolean), and must reject a list expiring at exactly `now` (`<=`), whereas `isExpired`'s own bound is strict (`<`).

3. **Message consistency in `verify-parsed-credential.ts`**: left as-is, taking you up on the option to keep the diff smaller rather than touching a third file for a wording-only change.

Verified: `vitest run` on the full `@agentcommercekit/vc` package - 111/111 passing. `oxlint` and `oxfmt` (both scoped and `pnpm run check:format` repo-wide) clean.

PR #155, opened 2026-08-13
fix(ack-pay): valibot decimals validator rejects negative values instead of clamping

### LINKED ISSUE #147: bug: `paymentOptionSchema.decimals` uses `toMinValue` (clamps) instead of `minValue` (validates)

ygd58:
Fixes #147

The valibot version of `paymentOptionSchema` used `v.toMinValue(0)` on the `decimals` field — a **transform** that silently clamps a negative number up to `0`, rather than a validator that rejects it. This diverged from the zod version of the same schema (`z.number().int().nonnegative()`), which correctly rejects negative values.

**Fix:** switched to `v.minValue(0)`, valibot's validating counterpart, so a payment option with a malformed negative `decimals` value is rejected by both schema implementations instead of being silently "fixed" by the valibot one.

**Tests:** added a regression test asserting both schemas reject a negative `decimals` value and agree on valid (zero/positive) values.

**Changeset:** added (`@agentcommercekit/ack-pay`, patch).

Verified locally: `vitest run src/schemas/decimals.test.ts` (3/3 passing — confirmed it fails against the old `toMinValue` code and passes with the fix), `oxlint` and `oxfmt --check` clean.

---

**AI usage disclosure** (per AI_POLICY.md): this fix was developed with Claude (Anthropic) assistance — identifying the divergence between the valibot and zod schemas, writing the fix, writing the test, and verifying locally. I reviewed and understand the change: it's a one-word swap (`toMinValue` → `minValue`) that changes valibot's `decimals` validation from a silent clamp to a rejection, matching zod's existing behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Payment option decimal values now consistently reject negative numbers.
  * Zero and positive decimal values continue to be accepted across supported validation formats.

* **Tests**
  * Added coverage to verify consistent decimal validation behavior.

### DIFF
--- a/.changeset/decimals-clamp-fix.md
+++ b/.changeset/decimals-clamp-fix.md
@@ -1,0 +1,16 @@
+---
+"@agentcommercekit/ack-pay": patch
+---
+
+Fix `paymentOptionSchema`'s `decimals` field silently clamping negative
+values instead of rejecting them (valibot schema)
+
+The valibot version of `paymentOptionSchema` used `v.toMinValue(0)` on the
+`decimals` field, which is a **transform** that silently clamps a negative
+number up to `0` rather than a validator that rejects it. This diverged from
+the zod version of the same schema (`z.number().int().nonnegative()`), which
+correctly rejects negative values.
+
+Switched to `v.minValue(0)`, valibot's validating counterpart, so a payment
+option with a malformed negative `decimals` value is now rejected by both
+schema implementations instead of being silently "fixed" by the valibot one.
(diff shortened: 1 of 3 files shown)


venables (changes requested):
Overall looks good but looks to be failing the linter

ygd58:
Thanks for catching that - fixed. It was an import-order issue in the test file (`oxfmt` wants `valibot` before `vitest`), pushed the corrected formatting. `pnpm run check:format` passes locally now.

PR #157, opened 2026-08-13
docs(jwt): drop the incorrect alg curve-alias claim in createJwt

devorun:
## Summary

Updated per review: dropped the API-widening approach and reduced this to the one-line doc fix requested.

`createJwt`'s JSDoc claimed `alg` accepts `secp256k1`/`Ed25519` as aliases, but the code never resolved them and every caller passes a `JwtAlgorithm` directly. This removes the stale alias sentence so the docstring matches the code — no code, type, or test changes.

## AI usage disclosure

Per `AI_POLICY.md`: written with AI assistance (Claude Code). I reviewed the change and understand it.

## Summary by CodeRabbit

* **Documentation**
  * Simplified the documentation for JWT algorithm configuration.
  * Clarified that the algorithm defaults to `ES256K`.

### DIFF
--- a/packages/jwt/src/create-jwt.ts
+++ b/packages/jwt/src/create-jwt.ts
@@ -25,9 +25,7 @@ export interface JwtHeader extends Omit<JWTHeader, "alg" | "typ"> {
  * @param payload - The payload to create the JWT from
  * @param options - The options to create the JWT from
  * @param header - Optional header overrides
- * @param header.alg - The algorithm to use for the JWT. Accepts `secp256k1` and
- *  `Ed25519` as aliases for `ES256K` and `EdDSA` respectively. Defaults to
- *  `ES256K`.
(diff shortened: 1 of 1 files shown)


venables (changes requested):
Thanks for catching this. The actual bug is the stale JSDoc on `createJwt` claiming alg accepts curve aliases.  Every caller already passes a `JwtAlgorithm`. Widening the public alg type to `JwtAlgorithm | KeyCurve` is not the surface we would want. A one-line PR that deletes the alias sentence from the docstring would be welcome.

devorun:
Thanks! Done — reverted the type widening (and its tests/changeset) and reduced this to the one-line docstring deletion you suggested. `alg` still only accepts a `JwtAlgorithm`.

venables (approved):
Thank you for converging on the doc-only fix; this is exactly right.

PR #158, opened 2026-08-14
fix(ack-pay): enforce payment request expiresAt

MrFaruk0:
`PaymentRequest.expiresAt` is documented as the point after which a payment request becomes invalid, but `verifyPaymentRequestToken` only enforced the standard JWT `exp` claim.

Payment requests created through ACK-Pay carry `expiresAt` in the JWT payload without an `exp` claim, so a validly signed request with a past `expiresAt` was still accepted by the default verification path.

**Fix:** after parsing the payment request, reject it when `expiresAt` has passed and expiry verification is enabled. `verifyExpiry: false` continues to skip expiry checks, preserving the existing receipt-verification behavior where receipts may outlive the payment request.

**Tests:** added regression coverage for past and future `expiresAt` values, `verifyExpiry: false`, and verifying a receipt over an expired payment request.

**Changeset:** added (`@agentcommercekit/ack-pay`, patch).

Related to #120, but non-overlapping: that PR validates and normalizes `expiresAt` values at the schema level; this change enforces the parsed timestamp during verification.

Verified locally: `@agentcommercekit/ack-pay` tests (38/38), package build, `oxlint`, `oxfmt --check`, and `git diff --check`.

---

**AI usage disclosure** (per AI_POLICY.md): this fix was developed with AI assistance (Codex - GPT 5.6 Sol).  I reviewed the final diff and understand the change.

## Summary by CodeRabbit

- **New Features**
  - Added expiration validation for payment requests when expiry verification is enabled.
  - Expired payment requests are rejected, while future-dated requests are accepted.
  - Expiry checks can be disabled when needed.
  - Payment receipts referencing expired requests continue to validate successfully.

### DIFF
--- a/packages/ack-pay/src/verify-payment-request-token.ts
+++ b/packages/ack-pay/src/verify-payment-request-token.ts
@@ -12,7 +12,7 @@ interface ValidatePaymentRequestTokenOptions {
    */
   resolver?: Resolvable
   /**
-   * Whether to verify the expiry of the payment request token
+   * Whether to verify JWT `exp` and PaymentRequest `expiresAt`
    */
   verifyExpiry?: boolean
   /**
@@ -56,6 +56,15 @@ export async function verifyPaymentRequestToken(
     throw new InvalidPaymentRequestTokenError(
       "Payment Request token is not a valid PaymentRequest",
     )
+  }
+
+  // `expiresAt` is a business-level expiry distinct from JWT `exp`.
+  if (
+    options.verifyExpiry !== false &&
+    output.expiresAt !== undefined &&
+    Date.parse(output.expiresAt) <= Date.now()
+  ) {
+    throw new InvalidPaymentRequestTokenError("Payment request has expired")
   }
 
   return {
--- a/.changeset/ack-pay-expires-at-verification.md
+++ b/.changeset/ack-pay-expires-at-verification.md
@@ -1,0 +1,5 @@
+---
+"@agentcommercekit/ack-pay": patch
+---
+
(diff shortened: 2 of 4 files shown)


packages/ack-pay/src/verify-payment-request-token.ts
```diff
@@ -58,6 +58,15 @@ export async function verifyPaymentRequestToken(
     )
   }
 
+  // `expiresAt` is a business-level expiry distinct from JWT `exp`.
+  if (
+    options.verifyExpiry !== false &&
+    output.expiresAt !== undefined &&
+    Date.parse(output.expiresAt) <= Date.now()
```
venables, line 65: The did-jwt library includes a skew variable to allow for clock drift, which we may want to support here? (defaults to `300` https://github.com/decentralized-identity/did-jwt/blob/master/src/JWT.ts#L522C97-L522C105)

venables (changes requested):
This change causes a downstream error in [examples/issuer/src/routes/receipts.ts:221-226](https://github.com/agentcommercekit/ack/blob/b0b61c5ee04f5c53032e67958022ad654e4ca251/examples/issuer/src/routes/receipts.ts#L221-L226): The receipt revoke route calls `verifyPaymentRequestToken` with only `resolver`, so after this PR it throws `InvalidPaymentRequestTokenError` for any receipt whose request `expiresAt` has passed. The route only needs `parsed.issuer`. 

Likely have to pass `verifyExpiry: false` in that call, the same way `verifyPaymentReceipt` does.

PR #159, opened 2026-08-14
fix(ack-pay): validate string payment amounts

Mabolla:
## Summary

Tighten `paymentOptionSchema.amount` validation so string amounts follow the same positive-integer contract as numeric amounts.

Currently the numeric branch requires a positive integer, while the string branch accepts any string. This allows values such as `""`, `"0"`, `"-1"`, `"1.5"`, and `"abc"` through schema validation even though payment amounts are expressed as integers in the smallest currency unit.

## Changes

- validate string amounts as positive integer strings in both Valibot and Zod schemas
- keep string support for integers beyond JavaScript's safe integer range
- add shared regression coverage for accepted and rejected values across both schema implementations
- add an `@agentcommercekit/ack-pay` patch changeset

## Validation

The regression tests cover positive numeric/string integers and reject zero, negative, fractional, empty, and non-numeric values in both schema implementations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Payment amounts now reject invalid values such as zero, negatives, fractions, empty strings, and nonnumeric text.
  * Large positive integer amounts provided as strings remain supported, including values beyond JavaScript’s safe integer range.

* **Tests**
  * Added coverage to verify consistent validation across supported schema implementations.

* **Chores**
  * Scheduled a patch release for the payment package.

### DIFF
--- a/.changeset/clean-amount-strings.md
+++ b/.changeset/clean-amount-strings.md
@@ -1,0 +1,5 @@
+---
+"@agentcommercekit/ack-pay": patch
+---
+
+Reject malformed, fractional, zero, and negative string payment amounts while preserving positive integer strings for values larger than JavaScript's safe integer range.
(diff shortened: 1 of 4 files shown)


venables (approved):
Thanks @Mabolla !

PR #160, opened 2026-08-15
fix(keys): reject invalid private JWK fields

burak33bb:
## Summary

- Require a present JWK private key field (`d`) to be a non-empty string
- Add regression coverage for invalid `d` values on Ed25519 and secp256k1 JWKs
- Add a patch changeset for `@agentcommercekit/keys`

## Why

The JWK guards distinguish public keys by the absence of `d` and private keys by its presence. This tightens that boundary so malformed private-key fields are not accepted by `isJwk` or `isPrivateKeyJwk`.

## Verification

- `pnpm --filter @agentcommercekit/keys test`
- `pnpm --filter @agentcommercekit/keys build`
- `node_modules\.bin\oxfmt.CMD --check packages/keys/src/encoding/jwk.ts packages/keys/src/encoding/jwk.test.ts .changeset/harden-private-jwk-guard.md`
- `git diff --check`

## AI usage disclosure

AI-assisted with ChatGPT/Codex for repository navigation, implementation, and tests. I reviewed and understand the final change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON Web Key validation to reject private keys with empty or non-string private-key values.
  * Corrected private-key detection for supported Ed25519 and secp256k1 keys.

* **Tests**
  * Added coverage for numeric and empty private-key fields across JWK validation checks.

* **Release**
  * Prepared a patch release for the keys package.

### DIFF
--- a/packages/keys/src/encoding/jwk.ts
+++ b/packages/keys/src/encoding/jwk.ts
@@ -6,6 +6,10 @@ function isRecord(value: unknown): value is Record<string, unknown> {
   return typeof value === "object" && value !== null
 }
 
+function hasValidPrivateKey(jwk: Record<string, unknown>): boolean {
+  return !("d" in jwk) || (typeof jwk.d === "string" && jwk.d.length > 0)
+}
+
 /**
  * JWK-encoding
  */
@@ -96,7 +100,7 @@ function isJwkSecp256(
     return false
   }
 
-  return true
+  return hasValidPrivateKey(jwk)
 }
 
 /**
@@ -142,7 +146,7 @@ export function isJwkEd25519(jwk: unknown): jwk is JwkEd25519 {
     return false
   }
 
-  return true
+  return hasValidPrivateKey(jwk)
 }
 
 export function isJwk(jwk: unknown): jwk is Jwk {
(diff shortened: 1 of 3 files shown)


packages/keys/src/encoding/jwk.ts
```diff
@@ -6,6 +6,10 @@ function isRecord(value: unknown): value is Record<string, unknown> {
   return typeof value === "object" && value !== null
 }
 
+function hasValidPrivateKey(jwk: Record<string, unknown>): boolean {
```
venables: Small nit-pick, but this doesn't check if it has a valid Private Key, this checks if it has a valid private key field.  May be worth changing to `hasValidPrivateKeyField` or someilar?

venables (changes requested):

burak33bb:
Addressed the review feedback: the helper is now named `hasValidPrivateKeyField`, and regression coverage includes both Ed25519 and secp256k1 invalid `d` values. Rebased onto the latest main and checks pass. @venables could you take another look when you get a chance?

venables:
Reopening this one. The review feedback was addressed and the fix is correct; apologies for the silence after your update. Running CI now.

venables (approved):
Thank you for the fix and for addressing the review feedback; sorry it sat after your update. Merging.

PR #161, opened 2026-08-16
fix(keys): accept a bare 0x prefix in isHexString

erhnysr:
## Problem

The JSDoc for `isHexString` at `packages/keys/src/encoding/hex.ts:41` documents:

```ts
isHexString("0x") // true
```

The implementation returns `false`. It strips the `0x` prefix and tests the remainder against `/^[0-9A-Fa-f]+$/` — the `+` quantifier requires at least one character, so `"0x"` reduces to `""` and fails.

Documented behaviour the code doesn't implement.

## Fix

An explicit branch for the bare-prefix case, so `"0x"` matches the documented result.

`""` deliberately stays `false`. The JSDoc says nothing about a bare empty string, and widening the quantifier to `*` would have changed that too — an undocumented behaviour change riding along with a documented fix. The explicit branch keeps the diff to exactly what the doc promises.

`isHexString` has no internal callers (verified by grep), so blast radius is limited to external consumers relying on the documented contract.

## Tests

A case in `hex.test.ts` covering the bare prefix, plus one pinning `""` as `false` so the distinction is intentional rather than incidental. Verified failing-first: before the change the `"0x"` case fails.

`@agentcommercekit/keys` — 101 tests pass. Full `pnpm run check` green: build 8/8, lint 0 warnings 0 errors, format clean, 29/29 tasks.

## Changeset

`patch` for `@agentcommercekit/keys`, matching the bump used by comparable one-line fixes such as #116.

---

AI Disclosure: Written with Claude Code. It located the doc/runtime divergence and wrote the guard and tests; I reviewed the diff and can explain the change and its trade-offs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated hexadecimal string validation to recognize both `"0x"` and `"0X"` as valid zero-length hexadecimal values.
  * Hexadecimal prefixes are now handled case-insensitively.
  * Empty unprefixed strings remain invalid.
  * Validation for non-empty hexadecimal values is unchanged.

* **Tests**
  * Added coverage for uppercase prefixes and prefixed and unprefixed empty hexadecimal strings.

### DIFF
--- a/.changeset/is-hex-string-empty-body.md
+++ b/.changeset/is-hex-string-empty-body.md
@@ -1,0 +1,5 @@
+---
+"@agentcommercekit/keys": patch
+---
+
+`isHexString` now returns `true` for a bare `0x` prefix with an empty body, matching its documented behavior. The check required at least one hex digit after the prefix, so `isHexString("0x")` returned `false` even though the JSDoc example states it returns `true`. A bare empty string with no prefix (`""`) continues to return `false`.
--- a/packages/keys/src/encoding/hex.test.ts
+++ b/packages/keys/src/encoding/hex.test.ts
@@ -44,6 +44,14 @@ describe("isHexString", () => {
     expect(isHexString("1234567890abcdef")).toBe(true)
   })
 
+  test("returns true for a bare 0x prefix with an empty body", () => {
+    expect(isHexString("0x")).toBe(true)
+  })
+
+  test("returns false for an empty string with no prefix", () => {
+    expect(isHexString("")).toBe(false)
+  })
+
   test("returns false for invalid hex strings", () => {
     expect(isHexString("0x1234567890abcdefg")).toBe(false)
     expect(isHexString("not hex")).toBe(false)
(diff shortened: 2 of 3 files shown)


erhnysr:
@venables good call on settling this in one place rather than case-by-case. Pushed a commit that folds in the uppercase-prefix case from #198: `isHexString` now strips `0x`/`0X` case-insensitively (`value.toLowerCase().startsWith("0x")`), matching `hexStringToBytes`, which already normalized case before checking the prefix — the two functions previously disagreed on this.

Added regression tests for `isHexString("0XABCDEF")` and the bare `isHexString("0X")` case (both now `true`), verified failing-first against the pre-fix code. Existing tests (lowercase prefix, no prefix, bare `0x`, empty string) are unchanged. Changeset updated to describe both the bare-prefix and case-insensitivity fixes together as one patch.

Full suite: 103/103 passing in @agentcommercekit/keys, `pnpm run check` clean (29/29 tasks, 0 lint/format errors). Ready for another look whenever you have a chance.

PR #163, opened 2026-08-16
Security & CI: Remove hardcoded secrets and harden workflows

mertcano:
This PR hardens the organization's CI pipelines by removing committed secrets from the check workflow.

Hardcoded Secrets Removed: Updated .github/workflows/check.yaml to remove committed issuer and verifier private-key literals, as well as the placeholder ANTHROPIC_API_KEY value.

Ephemeral Keys: The workflow now generates example signing keys ephemerally at runtime using the repository’s existing secret-generation helpers (examples/issuer/bin/secret and examples/verifier/bin/secret), exporting them strictly for the duration of the job.

## Summary by CodeRabbit

* **Chores**
  * Improved automated workflow security by preventing checkout credentials from being persisted and limiting sensitive credentials in the workflow environment.


venables:
Thanks again for this @mertcano , but please keep formatting out of the pull request.  In terms of the change, these values are intentionally visible so they can be tested. The PR body also mentions changes that re not in this pull request.

PR #166, opened 2026-08-16
fix(did,keys): emit a real Multikey for `publicKeyMultibase`

### LINKED ISSUE #165: bug(did): `publicKeyMultibase` on `Multikey` verification methods is not a Multikey
## Summary

`createDidDocument` emits verification methods with `type: "Multikey"`, but the `publicKeyMultibase` it puts on them is a multibase encoding of the **raw** public key bytes. A Multikey value has to be `multibase(base58-btc, multicodec-prefix ‖ key-bytes)`, so the value carries no algorithm identifier and a consumer cannot tell an Ed25519 key from a P-256 one.

`createDidKeyUri` in this same package builds the prefixed form correctly, so the two disagree about how the same key is encoded.

## Reproduction

Using the two keypairs already fixtured in `packages/did/src/methods/did-key.test.ts`:

```ts
const keypair = jwkToKeypair(jwk)

const doc = createDidDocumentFromKeypair({
  did: "did:web:example.com",
  keypair,
  encoding: "multibase", // "hex" and "base58" produce the same value
})

doc.verificationMethod[0].type // "Multikey"
doc.verificationMethod[0].publicKeyMultibase
createDidKeyUri(keypair)
```

| curve | `publicKeyMultibase` emitted | `createDidKeyUri()` on the same key |
| --- | --- | --- |
| Ed25519 | `z8myPWEuZj3T3WzBQu281AeLzE3pmU9h7em1YEGiD6ick` | `did:key:z6MknEES6VA14awWdV27ab5r1jtz3d6ct2wULmvU4YgE1wQ8` |
| secp256k1 | `zMi3oP598f15BeBYp3tdwtNEE3omWSogmoT8Q5ckb1N5AKKJj59KU61vMYYncCBDzGDaDkn7zib2ACobFPXB8TjGT` | `did:key:zQ3shNCcRrVT3tm43o6JNjSjQaiBXvSb8kHtFhoNGR8eimFZs`  |

Both columns encode the same key, so the multibase value on the verification method should equal the `did:key` method-specific identifier. Today it does not.

Two things differ:

1. **No multicodec prefix.** `encodePublicKeyMultibase` (`packages/keys/src/public-key.ts`) takes `curve` and ignores it, calling `bytesToMultibase(publicKey)` on the raw bytes. `createDidKeyUri` prepends `varint.encode(KEY_CONFIG[curve].multicodecPrefix)` before encoding. The same drop happens in `convertLegacyPublicKeyToMultibase` (`packages/did/src/create-did-document.ts`), which converts `hex` and `base58` inputs with a bare `bytesToMultibase(...)`.
2. **Uncompressed EC keys.** `generateKeypair` stores secp256k1/secp256r1 public keys uncompressed (65 bytes), which is why the secp256k1 row above is so long. `secp256k1-pub` (`0xe7`) and `p256-pub` (`0x1200`) identify the 33-byte compressed form — `createDidKeyUri` asks for compressed bytes and length-checks them against `KEY_CONFIG`.

The `createDidDocument` docblock already shows the prefixed form as the expected shape:

```ts
verificationMethod: {
  // ...
  publicKeyMultibase: "z6MkhaXgBZDvotDkL5257faiztiGiC2QtKLGpbnnEGta2doK"
}
```

`z6Mk…` is the Ed25519 Multikey form. The code produces `z8my…` for an Ed25519 key.

## Why it matters

`type: "Multikey"` is the Data Integrity Multikey type, and its `publicKeyMultibase` is defined as a multicodec-prefixed value. A relying party that resolves one of these `did:web` documents and follows that definition reads the leading bytes as an algorithm identifier and gets a wrong answer, or rejects the key.

This is reachable through the normal path, not just the `multibase` encoding: `hex` and `base58` are converted to a `Multikey` verification method too, and the repo's own tests exercise that (`did-resolver.test.ts`, `resolve-did.test.ts`, `did-web.test.ts` all pass `encoding: "hex"`). Only `encoding: "jwk"`, the default, is unaffected.

## Proposed fix

Make the multibase encoding of a public key produce a Multikey: compress secp256k1/secp256r1 keys, prepend the curve's multicodec prefix, then base58-btc multibase encode. That makes `publicKeyMultibase` equal the `did:key` identifier's multibase value for the same key, which is also a clean invariant to test.

The prefixes live in `KEY_CONFIG` in `packages/did/src/methods/did-key.ts` today, while the encoder that needs them is `encodePublicKeyMultibase` in `packages/keys`, which currently has no curve→multicodec mapping. So there is a placement decision here, and it changes the `publicKeyMultibase` every existing caller emits.

I have a working change and will open it as a draft PR against this issue, taking the route that keeps the mapping in `@agentcommercekit/keys` alongside the other curve metadata. Happy to move it, narrow it to the `did` package, or drop it entirely if you would rather handle the placement or the compatibility question differently — the measurements above are the part I wanted to get in front of you.

## AI assistance disclosure

Per the repository AI policy: this report was AI-assisted using Claude Code (Claude Opus). AI assistance was used to locate the inconsistency, produce the measurements above, and draft this issue. I reviewed the findings, can explain them, and take responsibility for what is reported here.

Dusk1e:
Fixes #165.

`createDidDocument` builds verification methods with `type: "Multikey"`, but the `publicKeyMultibase` on them was a multibase encoding of the raw public key bytes. A Multikey is `multibase(base58-btc, varint(multicodec code) ‖ key-bytes)`, so the value carried no algorithm identifier — a relying party reading it as a Multikey cannot tell an Ed25519 key from a P-256 one. `createDidKeyUri` already built the prefixed form, so the same key came out two different ways depending on which function produced it.

Using the two keypairs fixtured in `packages/did/src/methods/did-key.test.ts`:

| curve | before | after (= `createDidKeyUri`) |
| --- | --- | --- |
| Ed25519 | `z8myPWEuZj3T3WzBQu281AeLzE3pmU9h7em1YEGiD6ick` | `z6MknEES6VA14awWdV27ab5r1jtz3d6ct2wULmvU4YgE1wQ8` |
| secp256k1 | `zMi3oP598f15BeBYp3tdwtNEE3omWSogmoT8Q5ckb1N5AKKJj59KU61vMYYncCBDzGDaDkn7zib2ACobFPXB8TjGT` | `zQ3shNCcRrVT3tm43o6JNjSjQaiBXvSb8kHtFhoNGR8eimFZs` |

## Change

`publicKeyToMultikey(publicKey, curve)` in `@agentcommercekit/keys` prepends the curve's multicodec code as a varint and multibase-encodes the result. `encodePublicKeyMultibase` uses it, and `convertLegacyPublicKeyToMultibase` in `@agentcommercekit/did` routes `hex` and `base58` through the same encoder rather than calling `bytesToMultibase` directly — those two are converted to a `Multikey` verification method as well, so they carried the same defect.

The codes live in a new `keyCurveMulticodecs` map in `keys`, next to `keyCurves`. That is the placement decision I flagged in the issue: `keys` already owns curve metadata, key generation and every other encoding, and it is the package the encoder lives in. `KEY_CONFIG` in `did/methods/did-key.ts` is untouched, so nothing about `did:key` changes — instead the new test asserts the two agree, which is what keeps them from drifting apart again.

secp256k1 and secp256r1 keys are compressed to the 33-byte form their codes identify, via a new `compressPublicKey` on each curve module (`Point.fromBytes(...).toBytes(true)`, the same `@noble/curves` API `isValidPublicKey` already uses). `generateKeypair` stores these uncompressed, which is why the secp256k1 row above is so long. Ed25519 has a single 32-byte encoding and is passed through.

## Behaviour changes

- The `publicKeyMultibase` emitted by `createDidDocument`, `createDidDocumentFromKeypair`, `createDidWebDocument` and `encodePublicKey("multibase", ...)` changes for the `multibase`, `hex` and `base58` encodings. The default `jwk` encoding is unaffected. An already-published document keeps whatever it was published with; re-generating it produces the corrected value. Marked `minor` on both packages.
- `encodePublicKey("multibase", bytes, curve)` now throws for `secp256k1`/`secp256r1` when `bytes` is not a point on that curve, because compressing it means decoding it. This surfaced in `did-resolvers/did-resolver.test.ts`, which passed a 20-byte Ethereum address as a secp256k1 public key; that fixture is now a real key from `did-key.test.ts`. The test exercises the resolver cache, so the key material is incidental to it. Happy to make the compression lenient instead if you would rather this path keep accepting arbitrary bytes, though that would put non-keys back into `Multikey` values.

## Tests

- `keys`: the multibase encoding decodes to the curve's multicodec code followed by a key of the expected length (32 for Ed25519, 33 for the EC curves).
- `did`: for every curve, `publicKeyMultibase` equals the `did:key` identifier for the same key, across the `multibase`, `hex` and `base58` encodings — asserted against `createDidKeyUri`, which is independent of the code being changed.
- Two existing expectations that encoded the old value (`create-did-document.test.ts`, `methods/did-web.test.ts`) now build theirs with `publicKeyToMultikey`.

Both new tests fail on `main`. `keys` (102), `did` (73), `vc` (109), `ack-id` (39) and `ack-pay` (34) pass; `oxlint` and `oxfmt --check` are clean on the changed files.

One note on my local run: `packages/did/src/did-resolvers/pkh-did-resolver.test.ts` cannot run on Windows because the `did-pkh` fixture filenames contain `:`, which is #145. Unrelated to this change and left alone.

## AI assistance disclosure

Per the repository AI policy: this contribution was AI-assisted using Claude Code (Claude Opus). AI assistance was used to find the inconsistency, write the change and the tests, and run verification locally. I reviewed the final diff, can explain what it does and where it changes behaviour, and take responsibility for what is submitted here.

## Summary by CodeRabbit

* **New Features**
  * Public keys now use Multikey encoding compatible with `did:key` identifiers.
  * Added support for compressed secp256k1 and secp256r1 public keys.
  * Added curve-specific public-key compression and Multicodec mapping APIs.
* **Bug Fixes**
  * Invalid public keys are now rejected during multibase encoding.
  * Legacy hex and Base58 keys are converted using the correct Multikey format.
  * JWK output remains unchanged.

### DIFF
--- a/packages/keys/src/public-key.ts
+++ b/packages/keys/src/public-key.ts
@@ -102,11 +102,18 @@ function compressPublicKey(publicKey: Uint8Array, curve: KeyCurve): Uint8Array {
  * @param publicKey - The raw public key bytes
  * @param curve - The curve the key belongs to
  * @returns The Multikey string
+ * @throws If `publicKey` is not a valid public key on `curve`
  */
 export function publicKeyToMultikey(
   publicKey: Uint8Array,
   curve: KeyCurve,
 ): string {
+  // A Multikey says which curve its key is on, so bytes that are not a key on
+  // that curve must not get one of these prefixes.
+  if (!isValidPublicKey(publicKey, curve)) {
+    throw new Error(`Invalid ${curve} public key`)
+  }
+
   const compressed = compressPublicKey(publicKey, curve)
   const code = keyCurveMulticodecs[curve]
   const prefix = new Uint8Array(varint.encodingLength(code))
--- a/packages/keys/src/public-key.ts
+++ b/packages/keys/src/public-key.ts
@@ -1,3 +1,5 @@
+import { varint } from "multiformats"
+
 import * as ed25519 from "./curves/ed25519"
 import * as secp256k1 from "./curves/secp256k1"
 import * as secp256r1 from "./curves/secp256r1"
@@ -5,7 +7,7 @@ import { bytesToBase58 } from "./encoding/base58"
 import { bytesToHexString } from "./encoding/hex"
 import { publicKeyBytesToJwk, type PublicKeyJwk } from "./encoding/jwk"
 import { bytesToMultibase } from "./encoding/multibase"
-import type { KeyCurve } from "./key-curves"
+import { keyCurveMulticodecs, type KeyCurve } from "./key-curves"
 import type { Keypair } from "./keypair"
 
 /**
@@ -74,6 +76,50 @@ export function isValidPublicKey(
 }
 
 /**
+ * Compress a public key, for curves that have a compressed point encoding.
+ * Ed25519 keys have a single 32-byte encoding and are returned unchanged.
+ */
+function compressPublicKey(publicKey: Uint8Array, curve: KeyCurve): Uint8Array {
+  if (curve === "secp256k1") {
+    return secp256k1.compressPublicKey(publicKey)
+  }
+
+  if (curve === "secp256r1") {
+    return secp256r1.compressPublicKey(publicKey)
+  }
+
+  return publicKey
+}
+
+/**
+ * Encode a public key as a Multikey: the curve's multicodec code as a varint,
+ * followed by the key bytes, multibase-encoded with base58-btc.
+ *
+ * This is the value a `Multikey` verification method's `publicKeyMultibase`
+ * holds, and it is the same value that follows `did:key:` in a `did:key` URI
+ * for the same key.
+ *
+ * @param publicKey - The raw public key bytes
+ * @param curve - The curve the key belongs to
+ * @returns The Multikey string
+ */
+export function publicKeyToMultikey(
+  publicKey: Uint8Array,
+  curve: KeyCurve,
+): string {
+  const compressed = compressPublicKey(publicKey, curve)
+  const code = keyCurveMulticodecs[curve]
+  const prefix = new Uint8Array(varint.encodingLength(code))
+  varint.encodeTo(code, prefix, 0)
+
+  const prefixed = new Uint8Array(prefix.length + compressed.length)
+  prefixed.set(prefix)
+  prefixed.set(compressed, prefix.length)
+
+  return bytesToMultibase(prefixed)
+}
+
+/**
  * Convert a public key to a multibase string (used for DID:key)
  */
 function encodePublicKeyMultibase(
@@ -83,7 +129,7 @@ function encodePublicKeyMultibase(
   return {
     encoding: "multibase",
     curve,
-    value: bytesToMultibase(publicKey),
+    value: publicKeyToMultikey(publicKey, curve),
   }
 }
 
--- a/.changeset/multikey-public-key-multibase.md
+++ b/.changeset/multikey-public-key-multibase.md
@@ -1,0 +1,32 @@
+---
+"@agentcommercekit/keys": minor
+"@agentcommercekit/did": minor
+---
+
+`publicKeyMultibase` on a `Multikey` verification method is now an actual
+Multikey: `multibase(base58-btc, varint(multicodec code) ‖ key-bytes)`, with
+secp256k1 and secp256r1 keys in their 33-byte compressed form. It previously
+multibase-encoded the raw public key bytes, so the value carried no algorithm
+identifier and a relying party could not tell one curve's key from another's.
+`createDidKeyUri` already built the prefixed form, so the same key came out two
+different ways depending on which function produced it; the two now agree, and
+`publicKeyMultibase` equals the `did:key` method-specific identifier for the
+same key.
+
+This changes the value emitted by `createDidDocument`,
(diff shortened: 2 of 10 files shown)


packages/keys/src/public-key.ts
```diff
@@ -73,6 +75,50 @@ export function isValidPublicKey(
...
+ *
+ * This is the value a `Multikey` verification method's `publicKeyMultibase`
+ * holds, and it is the same value that follows `did:key:` in a `did:key` URI
+ * for the same key.
+ *
+ * @param publicKey - The raw public key bytes
+ * @param curve - The curve the key belongs to
+ * @returns The Multikey string
+ */
+export function publicKeyToMultikey(
```
venables, line 107: It looks like Ed25519 input is not validated in this method but EC input is. We could likely call `isValidPublicKey(publicKey, curve)` at the top w/ errors thrown for each curve

venables (changes requested):

packages/keys/src/public-key.ts
Dusk1e: Good catch, fixed sir

PR #176, opened 2026-08-24
fix(caip): allow hyphen in CAIP-2 namespace per spec

midasbal:
`caip2NamespacePattern` (`packages/caip/src/caips/caip-2.ts`) was `"[a-z0-9]{3,8}"`, but the [CAIP-2 spec](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-2.md) defines the namespace grammar as `[-a-z0-9]{3,8}`, hyphen included. The sibling `caip19AssetNamespacePattern` in `caip-19.ts` already uses the hyphenated form, and this file's own doc comment repeated the same wrong pattern, so the fix also corrects the comment for internal consistency. Added regression tests for a hyphenated namespace being accepted and an invalid character still being rejected, and a changeset for `@agentcommercekit/caip` (patch). One existing `did:pkh` resolver test's "invalid URI" fixture depended on the old regex rejecting a hyphenated namespace as its only source of invalidity; it's updated to use an uppercase namespace instead, which stays invalid for its own, unrelated, still-enforced reason.

**AI assistance disclosure:** This contribution was AI-assisted using Claude Code. AI was used to identify the spec discrepancy, implement the pattern/doc-comment fix, write the regression tests, and update the affected `did:pkh` test fixture. I reviewed the full diff and test output, understand the change and why the `did:pkh` test needed updating, and take responsibility for what's submitted here.

## Summary by CodeRabbit

- **Bug Fixes**
  - CAIP-2 chain identifiers now correctly accept hyphens in namespaces, in accordance with the specification.
  - Invalid underscore characters and uppercase namespace values continue to be rejected.

- **Tests**
  - Expanded validation coverage for hyphenated namespaces and invalid namespace formats.

- **Release**
  - Prepared a patch release for the CAIP package.

### DIFF
--- a/.changeset/curly-hyphens-parse.md
+++ b/.changeset/curly-hyphens-parse.md
@@ -1,0 +1,5 @@
+---
+"@agentcommercekit/caip": patch
+---
+
+Allow a hyphen in the CAIP-2 chain namespace, per spec (`namespace: [-a-z0-9]{3,8}`). `caip2NamespacePattern` previously omitted the hyphen, diverging from both the spec and the sibling `caip19AssetNamespacePattern`.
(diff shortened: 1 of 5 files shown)


venables (approved):
Thanks @midasbal !

PR #179, opened 2026-08-27
RFC: ACK-ID + ACK-Pay v2 

venables:
**Status: proposal.** Nothing here has shipped. The documents live under [`docs/ack-id/rfc/`](https://github.com/agentcommercekit/ack/tree/ack-id-core-rfc/docs/ack-id/rfc) and [`docs/ack-pay/rfc/`](https://github.com/agentcommercekit/ack/tree/ack-id-core-rfc/docs/ack-pay/rfc), unlisted from the docs site. We're asking for review of the design, not the prose.

## Summary

ACK-ID lets an AI agent prove three things to a relying party in one HTTP request: which agent is calling, which person or company is accountable for it, and whether that owner authorized this action. The request signature proves the agent, the grant's signature proves the owner, and the grant's scope proves the authorization.

This RFC proposes v2 as one small core plus six optional extensions:

- **Core**: identities (HTTPS URLs, spelled did:web in artifacts), key resolution (one fetch of `did.json`), grants (short-lived JWTs carrying audience, scope, and a possession pin), signed requests (RFC 9421, verifiable by Web Bot Auth infrastructure as deployed), a verification checklist, and revocation levers.
- **ext-controller**: ownership for unknown counterparties: controller derivation, control grants, and ownership proofs anchored in DNS or a code host. Drafted in full; this is the line from a request back to a legal entity.
- **ext-delegation, ext-revocation, ext-web, ext-attestations, ext-audit**: stubs that mark the split; normative text follows.

core-flow

ACK-Pay becomes a design language plus one normative profile: adopt the x402 offer-receipt artifacts (pinned at `x402-foundation/x402@59ac597`, the source shipped in `@x402/extensions` 2.15.0 and unchanged through 2.25.0), sign them with ACK-ID identities, and bind each receipt to the grant that authorized the payment. The trail runs receipt → agent → grant → owner → legal entity, and a third party can walk it with no callback to any participant.

trail

A core-only deployment (an org authenticating its own agents to its own services) is implementable with a stock JOSE library and an HTTP client in an afternoon. The whole setup, verified against `jose` as published:

```ts
import {
  calculateJwkThumbprint,
  exportJWK,
  generateKeyPair,
  SignJWT,
} from "jose"

// 1. Two keypairs: the owner signs grants, the agent signs requests
const owner = await generateKeyPair("EdDSA")
const agent = await generateKeyPair("EdDSA")

const agentJwk = await exportJWK(agent.publicKey)
agentJwk.kid = await calculateJwkThumbprint(agentJwk)
const ownerKid = await calculateJwkThumbprint(await exportJWK(owner.publicKey))

// 2. The agent's identity is a URL. Host this one static file
//    at https://acme.com/invoice-bot/did.json
const didDocument = {
  id: "did:web:acme.com:invoice-bot",
  verificationMethod: [
    {
      id: `did:web:acme.com:invoice-bot#${agentJwk.kid}`,
      type: "JsonWebKey2020",
      controller: "did:web:acme.com:invoice-bot",
      publicKeyJwk: agentJwk, // the only member core reads
    },
  ],
}

// 3. The owner mints a grant, pinned to the agent's key
const grant = await new SignJWT({
  scope: "invoices:read",
  cnf: { jkt: agentJwk.kid },
})
  .setProtectedHeader({ alg: "EdDSA", typ: "grant+jwt", kid: ownerKid })
  .setIssuer("did:web:acme.com")
  .setSubject("did:web:acme.com:invoice-bot")
  .setAudience("https://api.examplebank.com")
  .setIssuedAt()
  .setExpirationTime("15m")
  .setJti(crypto.randomUUID())
  .sign(owner.privateKey)

// Done. The agent signs each request (RFC 9421) and presents the
// grant in the Grant field. The verifier needs the same library
// plus one HTTPS GET of did.json.
```

## Key differences from v1

- **Plain JOSE replaces Verifiable Credentials.** Every artifact is a signed JWT verified with a stock JOSE library. The ideas VCs were for survive in core (self-issued identifiers, rotation that survives identity, portable signed claims, callback-free verification); the encodings (JSON-LD, the VC data model, presentation exchange) move out. A lossless grant-to-VC mapping lives in ext-attestations, so VC and eIDAS systems integrate as adapters rather than core dependencies.
- **One fetch replaces DID resolution.** The did:web spelling stays, and resolution is one fixed path: `<identity URL>/did.json`, or `/.well-known/did.json` for a bare domain, read for its `publicKeyJwk` entries only. No fallback on 404: failing closed is what makes key removal a revocation lever. The same document serves x402's key discovery.
- **Grants carry authority.** v1's controller credential asserts who controls an agent. A grant adds what the agent may do: `aud`, `scope`, `constraints`, `exp`, `jti`, and a required `cnf.jkt` possession pin. A copied grant is useless without the agent's private key; core has no bearer artifacts.
- **Requests carry the proof.** RFC 9421 HTTP message signatures replace v1's challenge-response exchange. Grants ride in a signed `Grant` header field, and the same signatures are verifiable by Web Bot Auth infrastructure already deployed at CDNs.
- **Verification is callback-free with explicit revocation levers.** RPs pin owner keys at onboarding and verify locally. Core revocation: short `exp`, single-use `jti`, agent key removal, owner key unpinning. List mechanisms move to ext-revocation, and an RP declares the longest grant lifetime it will accept.
- **Extensions replace conformance levels.** Unknown grant claims are ignored, and a claim that changes what a grant means must be named in `crit` (the RFC 7515 pattern applied to payload claims). A verifier that cannot evaluate a `crit` claim rejects the grant, so an extension claim can never downgrade a verifier that lacks the extension. Extensions also give their artifacts a second rejection surface (a distinct `typ`, a reserved scope token, an `aud` shape core rejects). Each extension is adopted and versioned on its own.
- **"Mandate" is renamed to "grant."** AP2 uses mandate for a human approving a purchase, and both artifacts can appear in one request. An AP2 mandate says a human approved a transaction; an ACK-ID grant says an owner authorized an agent to act.
- **Delegation chains link pairwise.** The `chain` claim carries one hash: the immediate parent. Ancestry is pinned transitively, decoy entries are structurally impossible, and the convention matches AP2's mandate chain, so grants slot into AP2/UCP flows. ext-delegation also names OAuth 2.0 Token Exchange (RFC 8693) as the intended mint carriage for issuance services, so token-exchange-shaped issuers (a Keycloak realm, a custody provider's signing API) can mint leaves through a standard interface while verification stays offline.
- **ACK-Pay adopts x402's artifacts.** Offers and receipts come from the x402 offer-receipt extension, pinned at a named version. v1's payment request token and VC receipts are superseded; the flow and role documents remain as non-normative design language. Card-network agent programs (Mastercard Agent Pay) and other rails map onto the same offer/receipt shape, as design language until a mapping is written.
- **Receipts name the accountable party.** A single `ack` member (`agent`, `grant`) inside the seller-signed receipt binds the payment to the grant that authorized it. A receipt without the binding stays a valid x402 receipt; it proves payment, and RPs that need the trail reject unbound receipts by policy.

## What didn't change

- The goal: a verifiable line from a request to the accountable entity behind an agent.
- Owner and agent are separate identities, and the owner is the accountable one.
- did:web is the spelling in artifacts; one hosted `did.json` serves ACK-ID, ACK-Pay, and x402.
- Human oversight in payments stays, as design language.

## Reading order

1. [`docs/ack-id/rfc/README.md`](https://github.com/agentcommercekit/ack/blob/ack-id-core-rfc/docs/ack-id/rfc/README.md): the split, settled decisions with rationale, document map
2. [`docs/ack-id/rfc/core.md`](https://github.com/agentcommercekit/ack/blob/ack-id-core-rfc/docs/ack-id/rfc/core.md): identities, keys, grants, signed requests, verification, revocation
3. [`docs/ack-id/rfc/ext-controller.md`](https://github.com/agentcommercekit/ack/blob/ack-id-core-rfc/docs/ack-id/rfc/ext-controller.md): controllers, ownership proofs, the full checklist
4. [`docs/ack-pay/rfc/core.md`](https://github.com/agentcommercekit/ack/blob/ack-id-core-rfc/docs/ack-pay/rfc/core.md): the x402 profile and the receipt binding

## AI usage

Claude Code drafted the review-response revisions and the thread replies; I reviewed every change.

### DIFF
--- a/docs/ack-id/rfc/core.md
+++ b/docs/ack-id/rfc/core.md
@@ -1,0 +1,473 @@
+# ACK-ID Core (draft RFC)
+
+**Status: proposal draft.** The key words MUST, MUST NOT, SHOULD, SHOULD NOT,
+and MAY are to be interpreted as described in RFC 2119.
+
+## 1. Scope
+
+ACK-ID Core defines how an agent proves who it is and what it may do. The
+proof is cryptographic and needs no callbacks. In core, the relying party
+already knows and trusts the owner that authorized the agent. Core defines:
+
+- identities and key resolution,
+- keys and thumbprints,
+- the grant artifact,
+- the signed-request carriage (Web Bot Auth compatible),
+- the verification checklist,
+- the revocation levers,
+- the extension mechanism.
+
+All other topics live in extension documents layered on this one: ownership
+discovery for unknown counterparties, delegation chains, revocation lists,
+the OAuth surface, attestations, and audit logs. A core-only deployment is
+complete and useful on its own. Example: an organization that authenticates
+its own agents to its own services. To serve unknown counterparties, and to
+trace a request back to an accountable entity, pair core with
+ext-controller.
+
+### 1.1 How it works
+
+Three parties. An **owner** (Acme) is accountable for an **agent** (Acme's
+invoice bot). A **relying party** (the API the bot calls) accepts the
+agent's requests because the owner authorized them.
+
+Two keypairs:
+
+- The agent holds a keypair and publishes the public key in one JSON
+  document at its identity URL (Section 3).
+- The owner holds a keypair and gives the public key to the relying party
+  once, at onboarding (Section 2).
+
+The flow:
+
+1. The owner signs a **grant**: a short-lived JWT saying this agent may
+   perform these actions at this relying party (Section 5).
+2. The agent signs each HTTP request with its own key (RFC 9421) and
+   attaches the grant in the `Grant` field (Section 6).
+3. The relying party verifies the request signature against the agent's
+   published keys (one HTTPS fetch). It verifies the grant against the
+   owner key it pinned at onboarding. It then checks that the grant's scope
+   covers the request (Section 7).
+
+Nobody calls anybody back. The owner can be offline. The agent proves
+possession of its key on every request. The relying party verifies
+everything locally except the one fetch of the agent's published keys.
+
+Every artifact verifies with a stock JOSE library. Conversion between an
+identity's two spellings (Section 3.1) is string manipulation.
+
+## 2. Terminology
+
+- **Relying party (RP)**: the service an agent calls. It verifies what the
+  agent presents. This document says **verifier** when a rule also applies
+  to parties that are not the target of a request (for example, an auditor
+  who walks an ACK-Pay receipt trail).
+- **Owner**: the accountable party behind an agent. In core, the RP knows
+  the owner out of band and pins its keys at onboarding.
+- **Agent**: software acting under an owner's authority.
+- **Identity**: an owner or agent, named by its HTTPS **identity URL** and
+  spelled as a did:web DID in artifacts (Section 3).
+- **Grant**: a JWT in which an issuer gives a subject bounded authority to
+  act (Section 5).
+- **Artifact**: any signed object this protocol defines. In core: grants
+  and signed requests.
+- **Onboarding**: the out-of-band step in which an RP learns an owner's
+  grant-signing public keys and records (**pins**) them against the
+  owner's identity. The channel belongs to the RP: a business agreement, a
+  dashboard, a config file. Core requires only that the channel exists and
+  supports unpinning (Section 8).
+
+## 3. Identity and key resolution
+
+### 3.1 Identities
+
+An identity has two equivalent spellings:
+
+- the **identity URL**: `https://acme.com` for an owner,
+  `https://acme.com/invoice-bot` for an agent. The URL is where the
+  identity's keys are published (Section 3.2).
+- the **did:web DID**: `did:web:acme.com`, `did:web:acme.com:invoice-bot`.
+  This spelling appears inside grants, as `iss` and `sub`.
+
+Conversion between the two is string manipulation: take the host, then the
+path segments, and join them with `:` in the DID spelling. The mapping
+works in both directions and gives exactly one result each way. One
+identity, two spellings, each used only where this document says so. The
+DID spelling exists for compatibility with adjacent layers that name
+signers as DIDs; the x402 offer-receipt extension resolves its signers'
+keys from the same document Section 3.2 defines. The URL spelling appears
+where an external spec requires a URL (`Signature-Agent`, per Web Bot
+Auth, Section 6).
+
+A host may carry any number of DNS labels:
+`https://agent.company.service.com` is a legal root identity, spelled
+`did:web:agent.company.service.com`, with keys resolved per Section 3.2
+rule 2. Host labels carry no relationship: core treats identities on
+different hosts as unrelated, whatever labels they share. The
+agent-to-owner binding is the grant (Section 5), and its `iss` may name an
+owner on any host the RP has onboarded. Controller derivation
+(ext-controller) operates on path segments only, never on host labels. A
+deployment that must choose between subdomain-per-agent and path-per-agent
+layouts should read that document's Section 2.
+
+Before a verifier uses an identity for comparison or URL construction, it
+MUST reject the identity if its URL form contains any of: a userinfo
+component, a port, a query, a fragment, a trailing slash, a dot-segment
+(`.` or `..`), an empty path segment, percent-encoding, or `.well-known`
+as its first path segment. The `.well-known` rule closes an aliasing hole.
+The path identity `https://acme.com/.well-known` would publish its keys at
+`https://acme.com/.well-known/did.json`. That is the same location that
+serves the root identity `https://acme.com`, so one document would answer
+for two identities. To compare two identities, a verifier MUST map both to
+one spelling, lowercase the host (and the scheme, in the URL form), and
+compare byte-for-byte. No other normalization is defined.
+
+### 3.2 Key resolution
+
+Given an identity, a verifier resolves its current public keys from its
+**DID document**, fetched from exactly one location (the did:web resolution
+rule):
+
+1. **Path identities** (URL form has a path): `<identity URL>/did.json`.
+2. **Root identities** (bare origin): `https://<host>/.well-known/did.json`.
+
+A verifier reads exactly one thing from the document: the key set. A key is
+a `verificationMethod` entry whose `publicKeyJwk` member carries a JWK (RFC
+7517). Core ignores every other member (`controller`, service endpoints,
+non-JWK key encodings); ext-controller assigns them meaning. A document
+with no parseable `publicKeyJwk` entries is malformed.
+
+A missing or malformed document ends resolution with failure. There is no
+fallback location. That strictness is what makes key removal a revocation
+lever (Section 8): a resolver that tried another location on 404 would
+quietly widen a key set its publisher deleted. Orgs whose keys live behind
+an OIDC `jwks_uri` use the opt-in discovery profile in ext-web.
+
+Non-normative: the same document serves the x402 offer-receipt extension's
+JWS key discovery, which resolves did:web signers from `did.json`. One
+hosted document covers ACK-ID verification and ACK-Pay artifact
+verification.
+
+Resolution rules:
+
+- All fetches MUST use HTTPS, MUST time out, and MUST cap response size.
+  Fetchers MAY follow redirects to a bounded depth (RECOMMENDED limit 3).
+  Every hop MUST use HTTPS and MUST pass the address checks below.
+- The identity URL is attacker-influenced until a signature verifies.
+  Fetchers MUST NOT connect to loopback, private (RFC 1918), link-local,
+  or unique-local addresses, on any hop. Deployments that face hostile
+  input SHOULD resolve DNS once and pin the address for the connection.
+- Key currency is judged against a fresh read of the document. A verifier
+  MUST NOT use a read older than 300 seconds and SHOULD cache no longer
+  than 60. Publishers MUST serve the document with cache lifetimes of 60
+  seconds or less. Removal of an agent key from the DID document then
+  invalidates everything that key signs within the cache bound: at most
+  300 seconds. (Owner keys are pinned, not resolved; their lever is
+  unpinning, Section 8.)
+
+## 4. Keys
+
+- **Algorithms.** A verifier MUST accept both `EdDSA` (Ed25519,
+  `kty: "OKP"`, `crv: "Ed25519"`) and `ES256` (P-256, `kty: "EC"`,
+  `crv: "P-256"`), and MUST reject any other `alg`. A verifier MUST select
+  the verification key by `kid` and MUST confirm the selected key's
+  `kty`/`crv` matches the asserted `alg`. Keys outside this set MUST be
+  ignored: a co-hosted DID document may carry RSA keys for other
+  protocols, and they play no part here. Issuers SHOULD issue Ed25519 by
+  default; P-256 exists for hardware-backed and FIPS-constrained keys.
+- **Custody.** Private keys are generated by the holder and MUST NOT be
+  transmitted to any server. The owner's key signs grants; each agent's key
+  signs that agent's requests. Owner keys SHOULD NOT live on the machines
+  running the agents they authorize.
+- **Rotation.** An identity holds one or more keys. Rotation is
+  add-then-remove and MUST NOT change the identity.
+
+### 4.1 Thumbprints
+
+Keys are named by their RFC 7638 JWK thumbprint, base64url encoded
+(computed per RFC 8037 for Ed25519). The thumbprint is the published key's
+name everywhere: the JWK `kid`, the `verificationMethod` fragment, the
+HTTP signature `keyid`, and the grant `cnf.jkt`. Lookup is by recomputed
+thumbprint. Given a `kid` or `keyid`, the verifier selects the resolved
+entry whose recomputed RFC 7638 thumbprint equals it. A stated JWK `kid`
+or fragment is a label, never the selector, so a published entry cannot
+claim another key's name. Exactly one entry may match. Zero matches, or
+more than one, is a rejection.
+
+## 5. Grants
+
+A grant is a JWT: header `alg` per Section 4, `typ: "grant+jwt"`, `kid` the
+issuer key's thumbprint. Verifiers MUST reject a grant whose `typ` differs.
+
+```json
+{
+  "iss": "did:web:acme.com",
+  "sub": "did:web:acme.com:invoice-bot",
+  "aud": "https://api.examplebank.com",
+  "scope": "invoices:read",
+  "constraints": { "region": "us" },
+  "iat": 1781035200,
+  "exp": 1781121600,
+  "jti": "grn_4kq8",
+  "cnf": { "jkt": "tH5Qw9..." }
+}
+```
+
+The example is non-normative; the recognized claims and their requiredness
+are:
+
+| claim         | requiredness | rule                              |
+| ------------- | ------------ | --------------------------------- |
+| `iss`         | REQUIRED     | DID of an onboarded owner         |
+| `sub`         | REQUIRED     | DID of the agent                  |
+| `aud`         | REQUIRED     | single string, exact match        |
+| `scope`       | REQUIRED     | space-delimited action scopes     |
+| `iat`, `exp`  | REQUIRED     | numeric; time rules below         |
+| `jti`         | REQUIRED     | unique per issuer                 |
+| `cnf`         | REQUIRED     | `jkt` possession pin              |
+| `constraints` | OPTIONAL     | open object; members only narrow  |
+| `nbf`         | OPTIONAL     | numeric; time rules below         |
+| `crit`        | OPTIONAL     | critical claim names; rules below |
+
+**Unknown claims are ignored; `crit` names the ones that must not be.** A
+verifier ignores any top-level claim it does not recognize. A claim that
+changes what the grant means MUST be named in `crit` by its issuer. A
+verifier MUST reject a grant whose `crit` names a claim the verifier does
+not implement and evaluate. This is the RFC 7515 `crit` pattern, applied
+to payload claims, and it is the extension mechanism. A claim like `chain`
+(ext-delegation) or `status` (ext-revocation) rides under `crit`, so a
+verifier that cannot evaluate it rejects the grant. `crit`, when present,
+MUST be a non-empty array of unique strings. It MUST NOT name any claim in
+the table above, and MUST NOT name a claim absent from the payload. A
+violation of any of these rules is a rejection.
+
+- `iss` and `sub` are DIDs (Section 3.1). `iss` MUST name an owner the RP
+  has onboarded (Section 2). The grant MUST verify against a key pinned
+  for exactly that owner (Section 7).
+- `aud` MUST be a single string exactly matching an identifier the RP
+  recognizes as its own. (Extensions define artifacts with other `aud`
+  shapes; core verifiers never accept those artifacts, by the rules above.)
+- `scope` is space-delimited. Every token is an action scope whose meaning
+  is owned by the audience, wildcard grammars included. The tokens
+  `control` and `register` are reserved for extensions (ext-controller and
+  ext-delegation define them). A core verifier MUST reject a grant that
+  carries either.
+- `constraints` is an object of audience-defined members. A member only
+  ever narrows what the grant allows: `{"region": "us"}` grants less than
+  no constraint at all. Because members only narrow, a verifier MUST
+  reject a grant that carries a member the verifier does not understand;
+  to skip one would widen the grant. This document assigns no meaning to
+  any member. Data that does not narrow authority does not belong in a
+  grant.
+- Time rules, with a clock-skew allowance `skew` (RECOMMENDED 60 seconds)
+  applied uniformly: `iat <= now + skew`; `nbf <= now + skew` when
+  present; `exp > now - skew`; `exp > iat`; and `exp > nbf` when `nbf` is
+  present. The last two compare issuer-set values against each other, so
+  no skew applies: an empty validity interval makes the grant malformed.
+  A malformed or non-numeric time claim is a rejection. (`nbf` is
+  recognized because some JOSE libraries add it by default.)
+- `jti` MUST be unique per issuer, with enough entropy that revocation and
+  single-use tracking by `jti` are well defined.
+- `cnf.jkt` pins the key that must prove possession at presentation. A
+  grant without it would be a bearer artifact, and core has no bearer
+  artifacts. Verifiers MUST reject a grant that lacks `cnf.jkt`. Verifiers
+  MUST match `cnf.jkt` against the possession key exactly, never against
+  "any current key of the subject."
+
+## 6. Signed requests
+
+Agents sign requests with RFC 9421 HTTP message signatures. The profile
+stays within what Web Bot Auth infrastructure verifies as deployed today,
+so one signature serves ACK-ID verifiers and WBA edges alike. The parts
+have three sources. RFC 9421 defines the signature itself and carries no
+credential. `Signature-Agent` comes from Web Bot Auth. The `Grant` field
+is defined by this document.
+
+- The signature MUST cover `@method`, `@target-uri`, `content-digest` when a
+  body is present, the `Signature-Agent` field, and the `Grant` field with
+  the `sf` parameter when grants are presented.
+- `Signature-Agent` carries the agent's identity URL (the URL spelling of
+  its DID, Section 3.1; Web Bot Auth requires a URL here). It is an
+  untrusted hint until the signature verifies. Because it is under the
+  signature, a middlebox cannot swap it.
+- `keyid` is the signing key's thumbprint. `created` (and `expires` where
+  used) bound replay. Verifiers MUST enforce a maximum `created` age
+  (RECOMMENDED default 300 seconds); the maximum is the RP's
+  replay-exposure ceiling, a policy choice like the Section 8 lifetime
+  maximum. When `expires` is present, it MUST be covered by the signature,
+  and a verifier MUST reject a request past it (the Section 5 `skew`
+  applies). `expires` only ever narrows the window: an `expires` beyond
+  the maximum `created` age does not extend acceptance. A captured request
+  replays verbatim within that window. RPs that accept non-idempotent
+  requests MUST therefore enforce replay protection: a cache over the
+  signature value, or the RFC 9421 `nonce` parameter.
+- `Grant` is a structured-field List of Tokens (RFC 9651), one token per
+  compact JWT. (An RFC 9651 Token must begin with an alphabetic character
+  or `*`. A compact JWT's first segment is base64url of a JSON header and
+  always begins `eyJ`, so grants satisfy the grammar.) A verifier MUST
+  reject a request whose `Grant` field fails structured-field parsing; it
+  MUST NOT salvage tokens from it. An unsigned `Grant` field is open to
+  middlebox changes. Coverage is what binds the grants to the request, and
+  a verifier MUST check that coverage, never assume it.
+- Each presented grant is evaluated independently; the request is authorized
+  only by grants that individually pass.
+
+Non-normative: a Web Bot Auth edge (for example, a CDN verifying crawler
+signatures) discovers keys per its own draft, at
+`/.well-known/http-message-signatures-directory` on the `Signature-Agent`
+origin, not via Section 3.2. An agent that wants those edges to recognize
+it also hosts that directory; see ext-web. The signature itself is the
+same either way.
+
+## 7. Verification
+
+The RP holds the owner's grant-signing public keys, pinned at onboarding
+(Section 2). The pinned set MUST contain grant-signing keys only; never,
+for example, an issuer's continuously online revocation key
+(ext-revocation). Three checks authorize a request. The only document
+fetch is the resolution of the agent's keys per Section 3.2.
+
+1. **Possession**: verify the request signature (Section 6) against the
+   agent's keys, resolved fresh per Section 3.2 from the `Signature-Agent`
+   identity URL, with the Section 4.1 thumbprint check against `keyid`.
+2. **Grant**, all of:
+   - `typ` is `grant+jwt`.
+   - The signature verifies against a key pinned for the owner named in
+     `iss`. Never merely "some pinned key": an RP with more than one
+     onboarded owner binds `iss` to the pinning entry, which stops one
+     owner minting grants attributed to another.
+   - `sub` equals the requesting agent's DID, mapped from the
+     `Signature-Agent` URL per Section 3.1.
+   - `aud` matches the RP exactly.
+   - The Section 5 claim table, `crit` rule, and time rules pass.
+   - The lifetime `exp - iat` does not exceed the RP's maximum (Section 8).
+   - No reserved scope tokens are present.
+   - `cnf.jkt` equals the possession key thumbprint.
+   - Every `constraints` member is understood and satisfied.
+   - The RP's Section 8 revocation posture holds.
+3. **Scope**: the grant's scope covers the requested action.
+
+Pinning is core's stance, and it is also core's boundary. An RP that must
+verify owners it never onboarded resolves issuer keys instead of pinning
+them; that is ext-controller's Section 6. Nothing else in the checklist
+changes.
+
+A passing verification proves continuity of key control and authorization.
+It never proves human presence, consent, or the legal identity behind a
+domain.
+
+## 8. Revocation
+
+Core's revocation story is complete against agent-key and grant
+compromise. It consists of these levers; the list mechanisms live in
+ext-revocation.
+
+- **Short `exp` is the primary mechanism.** Grant lifetimes SHOULD be
+  short (minutes to days). An `exp` on the order of minutes bounds
+  exposure with no revocation check at all.
+- **Single-use grants.** For one-shot authority, the RP records the `jti`
+  at first acceptance and rejects reuse; the grant is spent when used.
+  Which actions need single-use authority is the RP's call (a payment
+  authorization, a one-shot registration); core supplies the mechanism.
+  Redemption MUST be atomic: a check-and-set keyed by (`iss`, `jti`). A
+  separate read-then-write lets two concurrent presentations both pass.
+  The RP keeps the record until the grant's `exp` has passed.
+- **Key removal.** Removing an agent's keys from its DID document
+  invalidates everything the agent signs within the cache bound (Section
+  3.2).
+- **Owner-key unpinning.** Pinned owner keys are established out of band
+  at onboarding, and they are revoked the same way. RPs MUST support
+  unpinning a compromised owner key through the onboarding channel.
+  Unpinning invalidates every grant that key signed.
+
+A core RP MUST define a maximum acceptable grant lifetime, as a duration.
+It MUST reject any grant whose lifetime `exp - iat` exceeds it; the
+Section 7 checklist enforces this. Because acceptance also requires
+`iat <= now + skew`, the same bound caps remaining validity at acceptance.
+The maximum is the RP's exposure ceiling. An RP that needs to accept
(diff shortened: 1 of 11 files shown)


docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+## 1. Scope
+
+ACK-Pay Core defines how a payment leaves verifiable evidence: a **signed
+offer** the seller cannot later deny quoting, a **signed receipt** the buyer
+can present to third parties, and a binding from that receipt to the ACK-ID
+grant that authorized the payment. It defines no wire protocol, no
+settlement mechanism, and no new artifact formats. The artifacts are the
+x402 [offer-receipt extension](https://docs.x402.org/extensions/offer-receipt),
+adopted as published and profiled here. This profile pins the extension as
+shipped in `@x402/extensions` 2.22.0 (x402 repo, offer-receipt source at
```
EfeDurmaz16, line 15: At commit `59ac597` the implementation and docs disagree: `types.ts` defines `version, resourceUrl, scheme, network, asset, payTo, amount` and `transaction`, while the docs page says `offerType` and `txHash`, which Section 3 copies. The version is also off: package metadata at that commit says 2.15.0, not 2.22.0. Pick one pinned source of truth, fix the version, and state intentional deviations explicitly.

docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+`extensions` member; receipts in the `PAYMENT-RESPONSE` `extensions`), their
+field sets (`resourceUrl`, `offerType`, `network`, `amount`, `payTo`,
+`validUntil`; `resourceUrl`, `payer`, `network`, `issuedAt`, optional
+`txHash`), and their signature encodings are as defined by the x402
+offer-receipt extension. This profile constrains them:
+
+- **Signature scheme.** ACK-Pay conformance requires the JWS scheme. An
+  EIP-712/did:pkh signature MAY additionally be present; it carries no
+  ACK-Pay semantics and is passed through unevaluated.
+- **Signer identity.** The JWS signer MUST be a did:web identity per ACK-ID
```
EfeDurmaz16, line 50: Resolving the signer's keys proves who signed, not that they were entitled to. Nothing binds the signer DID to `resourceUrl`: anyone can host a did.json and sign offers for someone else's resource, and checks 1-4 pass. Needs an authorization rule: signer host matches the `resourceUrl` origin, or the resource names its accepted signer DIDs.

docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+accountable party. The binding that makes a receipt an accountability
+artifact:
+
+- When the paid request was ACK-verified (signed request plus grant, ACK-ID
+  core Sections 6-7), the seller SHOULD include in the receipt payload a
+  single profile-owned member, `ack`, carrying the agent's DID and the `jti`
+  of the grant presented:
+
+  ```json
+  "ack": { "agent": "did:web:acme.com:shopper", "grantId": "grn_4kq8" }
```
EfeDurmaz16, line 78: `jti` is only unique per issuer, and `ack` does not name the issuer, so a colluding owner can mint a grant with the same `sub` and `jti` and the trail walks to the wrong party. Bind by content: put the grant's hash in `ack` (base64url SHA-256 over the compact serialization, same construction as `chain`), or at least add the owner DID.

docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+
+1. **Keys**: resolve the signer's keys per ACK-ID core Section 3.2 from the
+   signer DID; thumbprint rules per ACK-ID core Section 4.1.
+2. **Signatures**: verify the JWS on the receipt, and on the offer when
+   present.
+3. **Matching**: receipt and offer agree on `resourceUrl` and `network`; the
+   offer's `validUntil` had not passed at `issuedAt`.
+4. **Freshness**: `issuedAt` is sane for the claimed transaction; where
+   `txHash` is present, it MAY be checked against the named network.
+5. **The trail**: when the `ack` member is present and the named grant is
```
EfeDurmaz16, line 118: Check 5 runs ACK-ID section 7 rules, but those are time-of-request checks. When a third party walks the trail later, the grant has expired and the seller may have rotated keys away, so every honest historical receipt fails. The profile needs a historical-verification mode: which checks run against the artifact's own timestamps, and the seller's key-publication obligation through the dispute horizon.

docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+   per ACK-ID core Section 7 rules: signature against the owner's keys,
+   `sub` equals `ack.agent`, `jti` equals `ack.grantId`. Where
+   legal-entity assurance is required, verify the owner's anchors per
+   ext-controller. A receipt whose `ack` member arrives without the grant
+   attributes the payment but proves no authorization; RPs that need the
+   trail treat it as unbound.
+
+What a passing verification proves: the named seller quoted these terms,
+acknowledged payment for this resource, and attributed the payment to this
+agent under this owner's grant. The receipt carries no amount or `payTo`.
```
EfeDurmaz16, line 130: Without an amount, a receipt proves payment happened but not what was settled, so the artifact caps out at fixed-price flows; upto/session/metered need "authorized X, settled Y <= X" and have nowhere to put Y. Open decision 1 already upstreams `ack`; propose attribution and outcome together: add `settledAmount`. x402's own scheme registry already defines `upto` and `deferred`, so these shapes are in scope for the pinned rail, not hypothetical.

docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+  key discovery: one hosted document serves both protocols. ACK-ID core's
+  resolution rules (one fixed location, SSRF protections, freshness)
+  apply.
+- **Algorithms.** EdDSA and ES256, per ACK-ID core Section 4. Whether ACK
+  verifiers also accept ES256K for wallet-adjacent deployments is an open
+  decision (Section 8).
+- **Key separation.** The offer/receipt signing key MUST NOT be the `payTo`
+  account and SHOULD NOT be an owner grant-signing key. A dedicated
+  signing key bounds what a compromise can mint.
+- **Offer expiry.** Offers MUST carry `validUntil`, and verifiers MUST
```
EfeDurmaz16, line 62: Nothing binds the authorization to a specific quote: any offer the seller presents within the grant's constraints is payable. Two fixes: define what a payment-authorizing grant is (no scope or amount semantics exist today), and for single-use grants allow pinning the offer (`constraints: {"offer": "<hash of the offer JWS>"}`), so a leaf authorizes exactly one signed quote and quote-swap disappears.

docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+   collision risk with upstream evolution, but upstreaming is still the
+   durable answer. Propose to the x402 extension either the `ack` member as
+   an extension point or native attribution fields, so non-ACK verifiers
+   learn to display the binding. This should be pursued now, while the
+   extension is young; the answer decides whether the member stays
+   namespaced or migrates to upstream-native fields.
+2. **ES256K.** The x402 JWS scheme allows secp256k1. Accepting it widens
+   wallet-key reuse; rejecting it keeps ACK-ID core's two-algorithm
+   discipline. Current draft: reject, revisit on deployment evidence.
+3. **Offer presentation.** Receipts verify without their offer; disputes
```
EfeDurmaz16, line 175: Buyer-side retention gap: pay, crash before storing the 200, and you have paid with no proof; re-issuance is undefined here and upstream. Sellers SHOULD re-issue idempotently for an already-settled payment, keyed by a stable settlement reference (`transaction` is optional), same content, `issuedAt` preserved.

EfeDurmaz16 (commented):
Read the full RFC set and cross-checked the payment profile against the pinned x402 source at commit 59ac597. The core + extensions split and the plain-JOSE posture feel right; comments ordered by severity.

**Open questions, non-blocking:**

- **ES256K (open decision 2):** rejecting has a real cost (EVM sellers must provision a non-wallet key) but I still lean reject: section 3 already forbids the signing key being `payTo`, and WebCrypto has no secp256k1, so accepting would make receipts unverifiable with native crypto on browser and edge runtimes.
- **did:jwks:** the SDK's did:jwks resolver uses the fallback chain core and ext-web forbid, and the RFC never mentions did:jwks. The README should say: superseded, folded into ext-web discovery, or profiled separately?
- **Standing authority:** do we want ext-delegation to profile recurring/metered payment flows? The intermediate-plus-leaf model is promising, but caps, cadence, sibling-leaf accounting, and buyer countersignatures need their own normative treatment. I would keep it out of this profile and track it as a follow-up design issue; happy to open one.
- **Stub drafting notes:** require every non-leaf ancestor to be `grant-int+jwt`; spell out the multi-level controller walk; name the revocation-key relationship; note that jti redemption and replay caches need coordination across an RP's acceptance domain.

Review assistance: Codex and Claude Code were used to cross-check the RFC against the pinned protocol sources and draft these notes. I reviewed the conclusions.

docs/ack-id/rfc/core.md
```diff
@@ -0,0 +1,473 @@
...
+
+Non-normative: the same document serves the x402 offer-receipt extension's
+JWS key discovery, which resolves did:web signers from `did.json`. One
+hosted document covers ACK-ID verification and ACK-Pay artifact
+verification.
+
+Resolution rules:
+
+- All fetches MUST use HTTPS, MUST time out, and MUST cap response size.
+  Fetchers MAY follow redirects to a bounded depth (RECOMMENDED limit 3).
```
EfeDurmaz16, line 154: Automatic redirect following in browser fetch, undici, and Workers never exposes hops; the per-hop checks only happen if the fetcher runs `redirect: "manual"` and walks the chain itself, which most will not do. Flip the default: refuse redirects, following as explicit opt-in for fetchers that implement the checks. We just made this change in the SDK resolver (#133).

docs/ack-id/rfc/core.md
```diff
@@ -0,0 +1,473 @@
...
+`did:web:agent.company.service.com`, with keys resolved per Section 3.2
+rule 2. Host labels carry no relationship: core treats identities on
+different hosts as unrelated, whatever labels they share. The
+agent-to-owner binding is the grant (Section 5), and its `iss` may name an
+owner on any host the RP has onboarded. Controller derivation
+(ext-controller) operates on path segments only, never on host labels. A
+deployment that must choose between subdomain-per-agent and path-per-agent
+layouts should read that document's Section 2.
+
+Before a verifier uses an identity for comparison or URL construction, it
```
EfeDurmaz16, line 113: The rejection list does not ban `:` inside a path segment, and colons are legal in URL paths, so `https://acme.com/a:b` and `https://acme.com/a/b` both map to `did:web:acme.com:a:b`. The mapping stops being bijective and one key document answers for two identities. Add path-segment `:` to the rejection list.

docs/ack-id/rfc/core.md
```diff
@@ -0,0 +1,473 @@
...
+## 6. Signed requests
+
+Agents sign requests with RFC 9421 HTTP message signatures. The profile
+stays within what Web Bot Auth infrastructure verifies as deployed today,
+so one signature serves ACK-ID verifiers and WBA edges alike. The parts
+have three sources. RFC 9421 defines the signature itself and carries no
+credential. `Signature-Agent` comes from Web Bot Auth. The `Grant` field
+is defined by this document.
+
+- The signature MUST cover `@method`, `@target-uri`, `content-digest` when a
```
EfeDurmaz16, line 286: Two enforcement gaps: covering `content-digest` binds the header, but nothing requires recomputing the digest over the received body, so a middlebox can swap the body under a valid signature. And `created` has a maximum age but no future bound; reject far-future values with the section 5 skew, like `iat`.

docs/ack-id/rfc/core.md
```diff
@@ -0,0 +1,473 @@
...
+- **Short `exp` is the primary mechanism.** Grant lifetimes SHOULD be
+  short (minutes to days). An `exp` on the order of minutes bounds
+  exposure with no revocation check at all.
+- **Single-use grants.** For one-shot authority, the RP records the `jti`
+  at first acceptance and rejects reuse; the grant is spent when used.
+  Which actions need single-use authority is the RP's call (a payment
+  authorization, a one-shot registration); core supplies the mechanism.
+  Redemption MUST be atomic: a check-and-set keyed by (`iss`, `jti`). A
+  separate read-then-write lets two concurrent presentations both pass.
+  The RP keeps the record until the grant's `exp` has passed.
```
EfeDurmaz16, line 374: Acceptance allows a grant until `exp + skew`, but the redemption record is kept only until `exp`. That leaves a skew-sized window where a spent grant passes again. Keep the record until `exp + skew`.

docs/ack-id/rfc/ext-controller.md
```diff
@@ -0,0 +1,314 @@
...
+### 5.2 Code-host anchor
+
+A file committed where only the org could commit it. One exact location per
+host, so two verifiers can never fetch different records. For GitHub: the
+repository named `.ack-id` under the owner, file `ack-id.json` at the
+repository root, read from the default branch through the host's raw
+endpoint (`https://raw.githubusercontent.com/<owner>/.ack-id/HEAD/ack-id.json`).
+Profiles for other hosts pin their equivalents. The shape is the same as
+the well-known file, with `anchor` = `github:<owner>` or the host's
+equivalent. The proof record MUST store the host's stable numeric owner ID
```
EfeDurmaz16, line 229: Three lifecycle gaps in ownership proofs. The expected numeric owner ID has no defined home: in the committed file it belongs to the rename-squatter it defends against, so it only works as trust-on-first-use, which should be stated. Proof claims have no `exp` and "whoever publishes proof status" is an undefined actor; either verifiers re-check anchors each time (say so) or a callback is back. And the DNS `jkt` pin is not wired into the Section 6 checklist: a verifier holding the pin can still accept artifacts signed by any other assertion key, defeating the pin against a compromised origin.

SkalorAI (commented):
Design review from the Skalor side. We implement against ACK-ID as an attestation issuer and run a pre-transaction control layer above the rails, so these are the places the RFC touches us directly.

Nothing here overlaps @EfeDurmaz16's comments, which I agree with — particularly content-binding `ack` and the need for a historical-verification mode.

Three of the seven are ext-attestations (carriage, subject binding through delegation chains, revocation); four are ACK-Pay (one reference rule, where a per-transaction control decision lives, issuer key obligations, rail-neutrality). Happy to go deeper on any of these in the thread.

docs/ack-id/rfc/ext-attestations.md
```diff
@@ -0,0 +1,67 @@
+# Extension: Attestations (stub)
+
+**Extracts v2 working draft section 7.**
+
+Third-party claims about an identity (KYB status, compliance level, audit
+results), distinct from grants: an attestation asserts, a grant authorizes.
+
+## Adds
+
+- **Carriage.** JWTs (`typ: "jwt"`) or SD-JWTs (`typ: "dc+sd-jwt"`), issued by
```
SkalorAI, line 10: **Attestations have no wire slot in a signed request.**

Core §6 makes the `Grant` field a List of grants, and core §7 rejects any token whose `typ` is not `grant+jwt`. An attestation (`typ: "jwt"` / `dc+sd-jwt`) therefore has nowhere to ride in the request itself. An RP that wants, say, an agent's bureau rating at decision time has to fetch it out of band — which reintroduces exactly the callback core removed.

Proposal: an `Attestation` structured field, same RFC 9651 List-of-Tokens grammar as `Grant`, covered by the request signature under the same coverage rule, each token evaluated independently under this extension. Core-only verifiers ignore the field entirely (it carries evidence, never authority), so it needs no `crit` entry and raises no downgrade concern.

If you'd rather not add a field, the alternative is to let `Grant` carry non-grant `typ`s and have core §7 skip rather than reject them — but that weakens the "every token in this field is a grant" invariant core currently relies on.

docs/ack-id/rfc/ext-attestations.md
```diff
@@ -0,0 +1,67 @@
...
+  any party under its own keys, typed by issuer-controlled `vct` per SD-JWT
+  VC. No central attestation-type registry.
+- **Subject binding** (the one normative rule): the attestation issuer MUST
```
SkalorAI, line 13: **Subject binding excludes exactly the deployment shape delegation exists for.**

Attestations bind to a subject identity, and ext-delegation's ephemeral did:key workers "die with their grant". Any attestation that is *behavioural* — a rating built from an agent's decision and settlement history — can never attach to a did:key leaf: the leaf has no history, and is gone before it earns any. Under the current text every orchestrated worker starts unattested.

Proposal: an attestation MAY name as `sub` any verified ancestor in the presented chain — typically the intermediate's `sub`, i.e. the durable agent or owner DID. A verifier that has walked the chain per ext-delegation then accepts an attestation whose `sub` equals any ancestor's `sub`, with `cnf` compared against that ancestor's key rather than the leaf's. The attestation follows the accountable identity rather than the throwaway key.

docs/ack-id/rfc/ext-attestations.md
```diff
@@ -0,0 +1,67 @@
...
+  verify, within its own session with the subject, that the subject controls
+  the identity being named (a session-bound challenge signed by a current key
+  of that identity). RPs SHOULD require issuers to document their binding
+  procedure.
+- **Selective disclosure.** SD-JWT only, and only here. Guidance: assert
+  results ("verified at level 2") rather than raw attributes (a date of
+  birth).
+- **Freshness.** Attestations SHOULD include `cnf` (an RFC 7638 thumbprint
```
SkalorAI, line 21: **Attestations have no revocation lever.**

ext-revocation gives grants a `status` claim under `crit`. Attestations have no equivalent, yet they are the artifact most likely to need pulling back inside their lifetime — a rating downgrade after an incident, a compliance status withdrawn.

SD-JWT VC already defines a `status` claim over the IETF Token Status List. Proposal: ext-attestations references it, and a verifier implementing this extension MUST check it when present, mirroring ext-revocation's obligation. Without that, a downgraded attestation stays good until `exp`, which pushes issuers toward minute-scale expiries and makes attestations useless as durable evidence.

docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+  verifiers also accept ES256K for wallet-adjacent deployments is an open
+  decision (Section 8).
+- **Key separation.** The offer/receipt signing key MUST NOT be the `payTo`
+  account and SHOULD NOT be an owner grant-signing key. A dedicated
+  signing key bounds what a compromise can mint.
+- **Offer expiry.** Offers MUST carry `validUntil`, and verifiers MUST
+  reject an offer past it. Sellers SHOULD keep offer validity short (the
+  x402 default of 300 seconds is a reasonable ceiling).
+
+## 4. Identity binding
```
SkalorAI, line 66: **Where does a per-transaction control decision live?**

The RFC has three artifact classes: a grant authorizes (owner), an attestation asserts (third party), a receipt acknowledges (seller). A per-transaction *control decision* from a third party — allow / refuse / pending, made against a specific offer under a specific grant, valid for seconds — fits none of them cleanly. We currently model it as an attestation carrying `cnf`, references to the offer and the grant, and a short `exp`.

Two asks:

1. Confirm attestation is the intended home, versus a `constraints` member on the grant. Core §9 reads as though the latter would also be legal: a clearance requirement *narrows*, narrowing semantics ship as `constraints`, and an unrecognized member is a rejection — which is the correct failure mode for a control requirement.
2. A design-language note that a receipt's `ack` MAY carry a reference to the clearance the seller relied on, so the walkable trail includes the control decision and not only the authorization.

Related: nothing in ACK-Pay records what was **refused**. Disputes ask "why was this paid" and "why was this blocked" in equal measure, and ext-audit's evidence bundles should admit control-decision artifacts for both outcomes.

docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+
+A receipt's `payer` is a settlement address, which moves money and names no
+accountable party. The binding that makes a receipt an accountability
+artifact:
+
+- When the paid request was ACK-verified (signed request plus grant, ACK-ID
+  core Sections 6-7), the seller SHOULD include in the receipt payload a
```
SkalorAI, line 73: **One reference rule for every cross-artifact link.**

+1 to @EfeDurmaz16 on binding `ack` by content rather than `jti`. Widening the point: the trail is about to grow several references of the same kind — receipt→grant (`ack`), grant→offer (the proposed `constraints.offer` pin), and any third-party artifact→{offer, grant, receipt} (an attestation citing what it evaluated).

Please define the reference construction **once**, in core, and reuse it: base64url SHA-256 over the compact serialization — the same construction as `chain`. For EIP-712 artifacts, define it over the EIP-712 typed-data digest, so one rule covers both formats the x402 extension admits.

One rule, one test vector, and every artifact that points at another verifies the same way.

docs/ack-pay/rfc/core.md
SkalorAI, line 118: **Historical verification needs an issuer-side obligation, not just a verifier mode.**

+1 to the historical-verification mode @EfeDurmaz16 raised. Adding the other half: whoever signs receipts, attestations or clearances that will be re-verified after key rotation needs a stated obligation to keep retired keys resolvable — or to be log-backed per ext-audit — through a declared dispute horizon.

We would commit to that as an attestation issuer and would rather it were a rule than a courtesy. Suggest the profile names the horizon explicitly, and makes ext-audit's event log the conformant way to satisfy it.

docs/ack-pay/rfc/core.md
```diff
@@ -0,0 +1,177 @@
...
+attests their satisfaction, and `txHash` (check 4) is the on-chain
+corroboration where present. It does not prove the resource was
+delivered or fit for purpose, that settlement is final on any particular
+rail, or that a human approved the payment.
+
+## 6. Other rails (design language)
+
+The offer/receipt pattern is rail-independent: an offer is a signed quote
+before payment, a receipt a signed acknowledgment after, whatever moved the
+money between them. Mappings for l402, card-network agent programs
```
SkalorAI, line 141: **Keep the attribution binding rail-neutral in shape.**

`ack` is defined as a member inside the x402 receipt payload. We clear the same policy across several rails, and the accountability trail matters identically on each.

Even as design language, it would help to name the binding as a standalone shape — an attribution object of `{agent, grant reference, rail, settlement reference}`, signed by whoever issues that rail's receipt — with the x402 profile being the case where it embeds in the receipt payload. Naming it now prevents four incompatible bindings later, and it is the shape a rail-neutral verifier actually consumes.

Small related question on ext-delegation's open item: for did:key subjects, is the intent that `Signature-Agent` is absent and the did:key is carried in the chain's leaf `sub` only, with the request key matched via `cnf.jkt`? We have implemented it that way and would like to confirm before it hardens.

aadopii (commented):
### Let a one-payment grant survive the 402 price challenge

There is an interaction between ACK-ID and ACK-Pay that I did not find in the existing reviews: when does checking a single-use payment grant become redemption of that grant?

A seller may authenticate an agent before showing a price. The agent sends a signed request, receives `402 Payment Required`, then retries with payment. Core §8 says the grant is spent at first acceptance. If middleware redeems it during the initial check, the paid retry fails as already spent even though the purchase has not started.

I propose making the ordering explicit:

1. Checking a payment grant and returning only a price challenge must not spend it.
2. On the paid retry, recheck the request, grant, and payment payload, then atomically and durably redeem (`iss`, `jti`) before settlement submission or the protected operation starts.
3. Only the winning request may start the authorized effects. A timeout after execution starts must not make the grant available for another operation.

This lets one purchase complete without asking the owner for another grant. It adds no token or header. A separately authorized paid quote can still consume its own grant.

The proposed changes are in https://github.com/agentcommercekit/ack/pull/215, targeting this RFC branch: a short ACK-Pay section, an ACK-ID cross-reference, and nine executable lifecycle examples integrated into the repository checks. The examples model ordering; they do not implement cryptography, distributed storage, or settlement. The full build and check commands pass in the fork's GitHub Actions run.

This concerns the stage before payment begins and is separate from the existing comments on spent-record expiry and receipt reissuance. Follow-ups could cover a production adapter and recovery using the payment rail's idempotency support.

AI assistance: I used Claude Code, Hermes, and Codex for source review and proposal development. Codex prepared the RFC edits, executable examples, and validation workflow, and ran the checks.

EfeDurmaz16:
A few additional thoughts. Keeping them in one comment to avoid crowding the review. These are design suggestions, not blockers; some build on the existing discussions.

1. **Constraints**
   a. Translating an ACK grant into wallet or rail-native authority must not broaden it. Required constraints that cannot be enforced should cause rejection.
   b. Could an optional profile support product/category-specific limits? For example, $100 per pair of shoes versus $400 per dataset license, alongside an aggregate budget. Classification must come from an owner-approved policy or trusted catalog, not an unchecked agent/seller label.

2. **Relationships and roles**
   a. A durable agreement could capture negotiated pricing, capacity, and service terms across short-lived grants. Referencing an agreement should not itself confer spending authority.
   b. Agent identity, customer account, fund owner, payment authority, fee sponsor, and relay are distinct roles. Recognizing an agent does not authorize access to a customer’s merchant account.

3. **After purchase**
   a. Refund, dispute, and order-access authority should survive the purchasing worker or grant. The authorized refund destination should not be inferred from the agent owner’s identity.
   b. Consider a separate fulfillment acknowledgment signed by an authorized recipient and bound to the receipt and delivered artifact. It would attest receipt, not quality; its absence would not prove nondelivery.

4. **Continuation and compatibility**
   a. Building on the existing `pending` discussion: human approval should resume the same operation, bind its current terms, and require fresh authorization where needed.
   b. Discovery should identify compatible extension versions and the active feature set. Negotiation must not silently drop required constraints.

5. **Future payment lifecycles**
   Building on the standing-authority discussion: subscriptions, sessions, `upto`, and batch settlement can remain follow-ups. Could we explicitly avoid assuming one grant = one request = one receipt = one settlement? Keeping today’s profile small should not make that relationship permanent.

docs/ack-pay/rfc/core.md
venables: Good catch, and thanks for checking against the actual commit rather than the page. The code is the source of truth now: Section 1 pins the offer-receipt source at `x402-foundation/x402@59ac597` (the last change to that directory; it first shipped in 2.15.0 and is unchanged through 2.25.0), and Section 3 uses `scheme` and `transaction` per `types.ts`. I also noted the two places the upstream docs page disagrees with the code so the next reader doesn't trip on it. cc63960

docs/ack-id/rfc/core.md
venables: Agreed, and #133 already made the SDK do this, so the spec was lagging the code. Core 3.2 now says fetchers MUST NOT follow redirects by default; following is opt-in and only for fetchers that walk each hop themselves. Section 10 and README decision 2 say the same. fc61a71

docs/ack-id/rfc/core.md
venables: Yep, that breaks the bijection. Rather than add `:` on its own I defined the segment alphabet: `A-Z a-z 0-9 . - _`, which is did:web's `idchar` set minus percent-encoding (already rejected). Anything else in a segment is a rejection, with your `a:b` example as the stated reason. 28afe7f

docs/ack-id/rfc/core.md
venables: Both taken. Verifiers MUST recompute `content-digest` over the received body and reject on mismatch, and `created` now has a future bound of the Section 5 `skew`, same as `iat`. 264f72a

docs/ack-id/rfc/core.md
venables: Right, that's a real window. Records are kept through `exp + skew` now, in core Section 8 and for registration grants in ext-delegation. a6f6214

docs/ack-pay/rfc/core.md
venables: Agreed, and it's worse than it looks since it moves accountability, which is the one thing the trail is for. `ack` now carries `grant`, an artifact reference: base64url SHA-256 over the compact serialization. The construction is defined once in ACK-ID core 4.2 (Skalor asked for one rule below) and `chain` reuses it. Check 5 compares the presented grant's reference to `ack.grant` before anything else. 76d5ff5

docs/ack-pay/rfc/core.md
venables: Done. ACK-ID core 4.2 defines the artifact reference once (base64url SHA-256 over the compact serialization, no padding), and `chain` and `ack.grant` both point at it. Any future cross-artifact link uses the same rule. I left EIP-712 out on purpose: the profile requires the JWS scheme and passes EIP-712 through unevaluated, so there's nothing on that side to reference yet. 76d5ff5

docs/ack-id/rfc/ext-attestations.md
venables: Agreed there's no slot today. ext-attestations now has a request-carriage bullet as a drafting requirement: a field with the same RFC 9651 grammar as `Grant`, same coverage rule, tokens evaluated independently, ignored by core-only verifiers since it carries evidence and never authority. I kept the `Grant` invariant intact rather than let non-grant `typ`s ride in it. I didn't name the field yet; that can wait for the normative pass. 2000280

docs/ack-id/rfc/ext-attestations.md
venables: Good point. A did:key leaf can't earn a rating before it dies. Added to ext-attestations as a drafting requirement: an attestation MAY name any verified ancestor in the presented chain as `sub`, with `cnf` compared against that ancestor's key. Normative text comes with the ext-delegation drafting pass. 2000280

docs/ack-id/rfc/ext-attestations.md
venables: Cheap and right. ext-attestations now references SD-JWT VC's `status` claim over the Token Status List (the same list ext-revocation uses for grants), and a verifier implementing the extension MUST check it when present. 2000280

docs/ack-pay/rfc/core.md
venables: Both taken. Section 6 now names the attribution as its own shape (agent DID, grant reference, settlement reference, signed by whoever issues the rail's receipt), with the x402 profile as the case where it embeds as `ack` and `transaction` is the settlement reference. 9ade080

On did:key: yes, that's the intended shape, and ext-delegation now says so. `Signature-Agent` absent, identity from the leaf's `sub`, key decoded from the identifier, `keyid` equal to its thumbprint and to `cnf.jkt`. Normative text is still a drafting requirement, but you're building the right thing. 0bd473f

docs/ack-pay/rfc/core.md
venables: On 1: an attestation, and I'd keep it there. A third party's clearance isn't the owner's authority, so it can't be a `constraints` member; constraints are set by the grant's issuer at issue time. What the owner can do is require one via a constraint (something like `constraints.clearance` naming the issuer it trusts), and the clearance itself rides as an attestation with `cnf`, the offer and grant references, and a short `exp`. That gives you the reject-on-unknown failure mode from the constraint side without putting third-party artifacts inside the grant.

On 2 and refusals: I'd rather keep `ack` to attribution for now and see where open decision 1 (upstreaming) lands before adding members to it. Refusal records are out of scope for the profile; ext-audit's evidence bundles are the natural home, and I'll carry it into that drafting pass.

docs/ack-pay/rfc/core.md
venables: You're right that the receipt caps out at fixed price today. Section 5 now says so explicitly, and open decision 1 proposes upstreaming a settled-amount field together with attribution, so `upto` and metered receipts can state what was paid. I'd rather it land upstream than as a profile member: it's a payment fact, and `ack` is for attribution. 66e32b0

docs/ack-pay/rfc/core.md
venables: Agreed the gap is real. I folded it into open decision 2 (offer presentation and receipt recovery) rather than add a SHOULD now. Idempotent re-issue is the right behavior, but it needs a request surface that neither the profile nor upstream defines, and the profile currently defines no wire protocol at all. 66e32b0

venables:
@EfeDurmaz16, on the open questions in your review:

- **ES256K**: closed as reject. Your WebCrypto point is now the stated reason in Section 3, plus the observation that key separation already forbids the signing key being `payTo`, so accepting it wouldn't buy wallet-key reuse anyway. 0a95426
- **did:jwks**: superseded. README decision 10 records it: the SDK resolver's jwks.json-then-OIDC fallback is exactly the chain core forbids, and ext-web's opt-in discovery profile covers the need. 3c06bdf
- **Standing authority**: agreed, keep it out of ext-delegation. Please open the follow-up issue; the intermediate-plus-leaf sketch is the right starting point.

On your follow-up list, two landed as one-liners in 9ade080. Section 1 says the profile describes one grant, one request, one receipt, one settlement, and doesn't fix that relationship (your 5). Section 6 says translation into rail-native authority never widens a grant, and an unenforceable constraint refuses the payment (your 1a). The rest (category limits, agreements, refund authority, fulfillment acks, discovery versions) are follow-up issues rather than profile text; happy to have you file them.

ak68a:
Coming at this from the downstream side.. I'm building a dispute evidence extension (#217) that depends on the v2 grant model and the receipt `ack` binding. A few things I've found while writing against these artifacts:

The content-hash `ack.grant` feels right.. The dispute evidence artifact embeds both the grant and the receipt, and a resolver confirms the receipt's `ack.grant` matches the SHA-256 of the embedded grant. Call it forgery defense I guess.. If the disputant forges a tighter grant, then the hash won't match what the receipt recorded at settlement time. The unified artifact reference rule from ACK-ID 4.2 extends cleanly to dispute→grant and dispute→receipt links with no new binding mechanism.

Also historical verification is critical for disputes because a dispute filed 60 days after a payment needs to verify the original grant (which has expired) and the original receipt (whose issuer may have rotated keys). The dispute evidence spec recommends a 90-day evidence window matching common chargeback windows. The historical verification mode plus the issuer-side obligation to keep old keys resolvable makes it viable imo, and without it any dispute filed after grant expiry is unverifiable.

On open decision #2 (offer retention) and receipt amount: the `scope-exceeded` reason code works by comparing the grant's `constraints.maxAmount` against the actual payment amount. Today that amount comes from the payment request embedded in the receipt. If the receipt carried the settled amount directly, the delta would be verifiable without depending on the embedded offer, which strengthens the evidence when the offer and the settlement diverge (partial fills, fee adjustments). Either way, offer retention is a hard requirement for dispute evidence.. Without the embedded payment request, the resolver can't verify amount-based deltas at all.

Full dispute evidence spec with verification checklist and working example is at [agent-payments-research/dispute-resolution](https://github.com/ak68a/agent-payments-research/tree/main/dispute-resolution).

ak68a:
A few observations from writing dispute evidence (#217) against these artifacts. Design suggestions, not blockers.

1. **The `ack` binding is SHOULD, not MUST.** ACK-Pay Section 4 lets a compliant seller omit it; the receipt then attributes payment to no one. Every downstream consumer of the trail (dispute evidence, audit bundles) breaks silently. If the profile's purpose is accountable agent commerce, the binding should be MUST within the profile.

2. **`constraints` has no shared vocabulary.** Section 5 assigns no meaning to any member. A third party walking the trail encounters `{"maxAmount": "10000"}` and cannot tell if that is subunits or major units without the RP's specific semantics. A small registry of well-known names with specified semantics, alongside the open audience-defined mechanism, would give extensions and verifiers something stable to work with.

3. **Single-use intent is not expressible by the issuer.** Section 8 makes single-use enforcement the RP's call. The owner cannot declare "this grant covers one payment." A well-known constraint `singleUse: true` would close it: the owner states the intent, and an RP that does not recognize the member rejects per the Section 5 unknown-constraint rule. The mechanism already enforces it; it just needs a name.

PR #181, opened 2026-08-28
fix(ack-pay): bind payment request token subject to its id

crazywriter1:
## Problem

`createPaymentRequestToken` always sets `sub` to the payment request id, and the create-side tests assert that. `verifyPaymentRequestToken` never checked the binding, so a validly signed token could carry a different `sub` and still verify.

That is the same class of gap #88 closed for receipts: the create path establishes a contract and the verify path did not enforce it.

## Fix

After parsing the payload with `paymentRequestSchema`, reject the token when the JWT `sub` is missing or does not equal the parsed payment request `id`.

## Tests

Added a regression test that signs a payment request with `sub` set to a different value and expects `InvalidPaymentRequestTokenError` with no `cause`, same pattern as the existing invalid-payload test.

Patch changeset included for `@agentcommercekit/ack-pay`.

## AI usage

Per [AI_POLICY.md](https://github.com/agentcommercekit/ack/blob/main/AI_POLICY.md), Cursor did the work: it found the create/verify asymmetry during a rescan of main, wrote the check and the test, and added the changeset. I read the diff before opening this.

## Summary by CodeRabbit

- **Bug Fixes**
  - Payment request tokens are now rejected when their subject does not match the payment request ID.
  - Invalid or mismatched token subjects now produce a clear verification error.

- **Documentation**
  - Clarified that JWT subject claims must match the payment request ID.

- **Tests**
  - Added coverage confirming mismatched token subjects fail verification.

- **Chores**
  - Prepared a patch release for the payment request package.


packages/ack-pay/src/verify-payment-request-token.test.ts
```diff
@@ -183,6 +183,37 @@ describe("verifyPaymentRequestToken", () => {
     expect(error.cause).toBeInstanceOf(Error)
   })
 
+  it("throws when sub does not match id without cause", async () => {
+    const mismatchedToken = await createJwt(
+      {
+        ...paymentRequest,
+        id: paymentRequest.id,
```
venables: Don't need this line since its coming from the spread above

venables (changes requested):
With this change, now the `paymentRequestToken` row in  [docs/ack-pay/payment-request-payload.mdx:73](https://github.com/agentcommercekit/ack/blob/main/docs/ack-pay/payment-request-payload.mdx#L73) does not say the JWT `sub` claim must equal `paymentRequest.id`; a third-party issuer that follows the docs now fails verification. 

Can you please add one sentence to that row that states the `sub` contract.

crazywriter1:
Thanks for the review sir. Good catch on both. Removed the redundant `id` in the test, and added a sentence to the `paymentRequestToken` row in `payment-request-payload.mdx` stating that the JWT `sub` claim must equal `paymentRequest.id`. @venables

crazywriter1:
Friendly ping. The earlier review notes are addressed (`id` line removed, `sub` contract documented in `payment-request-payload.mdx`). Happy to adjust anything else if needed. @venables

PR #183, opened 2026-08-28
fix(ack-pay): reject empty payment option identifiers

crazywriter1:
## Problem

`paymentOptionSchema` accepted empty strings for `id`, `currency`, and `recipient`, and `paymentRequestSchema` accepted an empty request `id`. Other fields already reject invalid values — `amount` must be positive, and the zod schema rejects negative `decimals` — so an all-empty option was not an intentional allowance, just `v.string()` / `z.string()` with no length check.

A signed payment request token can carry those fields through to verification, so the schema is the gate for HTTP 402 bodies and anything that parses them with `verifyPaymentRequestToken`.

## Fix

Introduce a shared `nonEmptyString` in both schema files and use it for the payment request `id` and for `id`, `currency`, and `recipient` on each payment option. Valibot and zod stay in sync, per the repo's dual-schema rule.

## Tests

Extended `schemas.test.ts` with parity coverage: each schema rejects a payment option with an empty `id`, `currency`, or `recipient`, and rejects a payment request with an empty top-level `id`.

Patch changeset included for `@agentcommercekit/ack-pay`.

## AI usage

Per [AI_POLICY.md](https://github.com/agentcommercekit/ack/blob/main/AI_POLICY.md), Cursor did the work during a rescan of main: it measured the empty-string acceptance, wrote the schema change and parity tests, and added the changeset. I read the diff before opening this.

## Summary by CodeRabbit

- **Bug Fixes**
  - Payment validation now rejects empty payment and request identifiers, currencies, and recipients.
  - Payment amounts provided as strings must be positive whole numbers, preventing empty, zero, zero-padded, or non-numeric values.

- **Tests**
  - Added coverage for invalid empty fields across supported payment validation formats.

- **Documentation**
  - Added release notes describing the updated validation behavior.


kutluhaneth46:
This PR is currently conflicting with `main`. Opening a fresh, rebased PR with the same empty-string rejection for payment request/option identifiers so the fix can land cleanly.

crazywriter1:
Conflicts with main are fixed.

PR #197, opened 2026-09-04
demo(payments): add a cumulative spend budget to the policy guard

### LINKED ISSUE #138: demo(payments): add a cumulative spend budget to the policy guard

kutluhaneth46:
## Summary
- Add an in-memory rolling-window spend ledger under `demos/payments` so the policy guard bounds cumulative spend, not only a per-transaction cap.
- Introduce `authorizePayment` on top of unchanged `evaluatePaymentPolicy`, with check-and-reserve as one synchronous step.
- Key reservations by payment request id + payment option id so the Stripe URL + callback path authorizes once; commit on receipt, release on failure.

Fixes #138.

## Notes
Everything stays in `demos/payments` (no package/protocol change). Budget breaches return `denied`, matching the existing per-transaction cap. Still demo-grade: in-memory, single-instance, denies rather than escalating to human approval.

## Test plan
- [ ] `pnpm --filter ./demos/payments exec vitest run`
- [ ] Confirm split-attack test denies the 4th payment at the window limit
- [ ] Confirm idempotent re-authorization does not double-count
- [ ] Confirm commit/release and window expiry behaviour

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added rolling-window cumulative spend limits alongside per-transaction caps.
  * Payment authorization now tracks spending by payer and currency, preventing transactions that exceed configured budgets.
  * Added retry-safe reservation handling for successful, failed, and repeated payment attempts.
  * Settled payments can continue through receipt issuance when they exceed the rolling budget.
  * Payment requests and receipts are signed using the payer identity.
  * Added verified payment settlement handling, expiration of abandoned settlements, and timeout protection for receipt retrieval.
* **Documentation**
  * Updated payment policy documentation with budget rules, approval requirements, and demo ledger limitations.

### DIFF
--- a/demos/payments/README.md
+++ b/demos/payments/README.md
@@ -20,7 +20,7 @@ This interactive command-line demo showcases a common use case: the **Server-Ini
   - Handling currency conversions.
   - Integrating compliance checks (KYC/AML).
   - Facilitating complex payment routing.
-  - Enforcing local payment policy before returning an execution URL or signing a receipt-service payload.
+  - Enforcing local payment policy, including a cumulative spend budget, before returning an execution URL or signing a receipt-service payload.
 
 You can learn more about the full ACK-Pay protocol at [www.agentcommercekit.com](https://www.agentcommercekit.com).
 
@@ -36,9 +36,10 @@ Payment Services should replace it with their own owner, risk, compliance, or
 human-approval system. The important safety boundary is that policy enforcement
 happens before execution or signing:
 
-- known low-value recipient: continue automatically
+- known low-value recipient, within budget: continue automatically
 - unknown recipient: return `approval_required`
-- amount above the illustrative per-transaction cap: deny before payment
+- amount above the per-transaction cap: deny before payment execution
+- amount that would exceed the rolling spend budget: deny before payment
   execution
 
 The per-currency cap is expressed in each currency's smallest subunit, so a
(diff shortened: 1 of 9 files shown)


kutluhaneth46:
﻿Addressed the CodeRabbit finding on settled Stripe callbacks:

- Callback re-auth now passes `allowOverBudget: true`
- Over-budget still runs per-transaction checks; on window breach it `recordOverBudget` and continues to receipt issuance instead of 403
- Added unit coverage in `payment-policy.test.ts` / `spend-ledger.test.ts` (25/25 green locally)

kutluhaneth46:
CodeRabbit reservation-release nitpick addressed on the `/` path: if payment URL creation fails after `enforcePaymentPolicy`, the spend reservation is released via `spendLedger.release(reference)` so the window budget is not held.

kutluhaneth46:
Addressed the latest CodeRabbit majors on this branch:

- **Stripe settlement before `allowOverBudget`:** the `/` path now issues a pending settlement ticket for the payment-attempt reference; `/stripe-callback` only sets `allowOverBudget: true` after consuming that ticket with a Stripe-shaped `evt_…` event id (demo stand-in for a signed webhook / Events API check). Unsolicited callbacks get `401`.
- **Receipt Service hang:** `fetchWithTimeout` (10s AbortController) so a stalled receipt call hits the existing `catch` and `spendLedger.release(reference)`.

Coverage: `stripe-settlement.test.ts` + existing policy/ledger suites — **33/33** green locally.

kutluhaneth46:
Addressed the latest CodeRabbit majors on this branch:

- **Signed Stripe events:** callbacks now require an HMAC (`metadata.signature` over `eventId.reference`) with the demo webhook secret — format-only `evt_…` ids are no longer enough for `allowOverBudget`.
- **Don't consume before receipt:** `verify()` marks settlement verified but retains it; `commit()` runs only after Receipt Service succeeds, so recoverable receipt failures can retry with the same event.
- **Expire abandoned pendings:** pending settlements prune after a 30m TTL (abandoned payment-URL lifetime).
- **Body-aware fetch timeout:** `fetchWithTimeout` keeps the AbortController armed through response body buffering, so a stalled JSON body still releases the spend reservation path.

Coverage: `stripe-settlement` / policy / ledger tests — **37/37** green locally.

venables:
Thank you for this. A cumulative budget for the payments demo is already in review in #187, which was opened first and follows earlier review feedback. Review comments on that PR are the most helpful way to move this forward. Closing as a duplicate of #187.

PR #198, opened 2026-09-04
fix(keys): accept uppercase 0X prefix in isHexString

kutluhaneth46:
## Summary
- `isHexString` only stripped a lowercase `0x` prefix, while `hexStringToBytes` already accepts `0X…`.
- Make the guard strip the prefix case-insensitively so values the decoder accepts also pass the type guard.
- Add a regression test for `0XABCDEF`.

This is distinct from #161 (bare `0x` / empty body); that PR does not make the prefix check case-insensitive.

## Test plan
- [ ] `pnpm --filter @agentcommercekit/keys exec vitest run src/encoding/hex.test.ts`
- [ ] Confirm `isHexString('0XABCDEF') === true`
- [ ] Confirm invalid hex still returns false

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Hex string validation now accepts both lowercase `0x` and uppercase `0X` prefixes, matching hex string conversion behavior.

* **Tests**
  * Added coverage for valid hexadecimal strings using the uppercase `0X` prefix.

### DIFF
--- a/.changeset/is-hex-string-uppercase-prefix.md
+++ b/.changeset/is-hex-string-uppercase-prefix.md
@@ -1,0 +1,5 @@
+---
+"@agentcommercekit/keys": patch
+---
+
+Accept an uppercase `0X` prefix in `isHexString`, matching `hexStringToBytes`.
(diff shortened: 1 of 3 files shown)


venables:
Thank you for the contribution. What isHexString should accept, including prefix variants, is already an open question on #161, and we would like to settle it once rather than widen the function case by case. Closing this one; feel free to add your case to that discussion.

PR #202, opened 2026-09-05
fix(ack-pay): reject negative payment option decimals

### LINKED ISSUE #147: bug: `paymentOptionSchema.decimals` uses `toMinValue` (clamps) instead of `minValue` (validates)

kutluhaneth46:
## Summary
- Valibot `paymentOptionSchema.decimals` used `v.toMinValue(0)`, which **clamps** negative values to `0` instead of rejecting them.
- Zod already uses `.nonnegative()`, so the two schema implementations disagreed.
- Switch Valibot to `v.minValue(0)` and add parity tests covering accept/reject cases.

Fixes #147

## Test plan
- [x] `pnpm --filter @agentcommercekit/ack-pay exec vitest run src/schemas/payment-option.test.ts`
- [x] Confirm `decimals: -1` fails in both Valibot and Zod
- [x] Confirm `decimals: 0 | 2 | 18` still succeeds

## AI usage disclosure
AI-assisted with Cursor for repo navigation, the schema change, tests, and changeset. I reviewed and understand the Valibot `toMinValue` vs `minValue` distinction and that this aligns Valibot with the existing Zod validator.

## Summary by CodeRabbit

- **Bug Fixes**
  - Negative decimal values in payment options are now rejected instead of being silently changed to zero.
  - Decimal values must be non-negative whole numbers, with valid values such as 0, 2, and 18 accepted.

- **Tests**
  - Added coverage for valid and invalid decimal values across supported validation formats.

### DIFF
--- a/.changeset/ack-pay-decimals-min-value.md
+++ b/.changeset/ack-pay-decimals-min-value.md
@@ -1,0 +1,7 @@
+---
+"@agentcommercekit/ack-pay": patch
+---
+
+Reject negative `decimals` in the Valibot payment option schema instead of clamping them to `0`.
+
+`paymentOptionSchema` used `v.toMinValue(0)`, which silently rewrites negative values. That diverged from the Zod schema (`z.number().int().nonnegative()`) and could hide invalid payment precision. Switch to `v.minValue(0)` so both schemas reject the same inputs.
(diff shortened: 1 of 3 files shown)


venables:
Thank you for the fix. This change is already open in #155, with earlier versions in #149 and #151. Closing as a duplicate of #155.

PR #203, opened 2026-09-05
fix(ack-pay): reject empty payment request and option fields

kutluhaneth46:
## Summary
- Payment option `id` / `currency` / `recipient`, payment request `id`, and receipt claim `paymentOptionId` previously accepted empty strings via bare `v.string()` / `z.string()`.
- Other fields already reject invalid values (e.g. `amount`), so blank identifiers were an inconsistency rather than an intentional allowance.
- Introduce a shared non-empty string helper in both Valibot and Zod schemas, with parity tests.

This is a clean rebase of the same fix attempted in #183 (currently conflicting with `main`).

## Test plan
- [x] `pnpm --filter @agentcommercekit/ack-pay exec vitest run src/schemas/schemas.test.ts`
- [x] Empty `id` / `currency` / `recipient` on a payment option are rejected by both schemas
- [x] Empty payment request `id` is rejected by both schemas

## AI usage disclosure
AI-assisted with Cursor for identifying the empty-string acceptance, implementing the shared non-empty string validators, tests, and changeset. I reviewed the dual-schema change and understand why these fields need length checks at the schema gate used by HTTP 402 bodies and token verification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Payment requests and options now reject empty identifiers, currencies, and recipient values.
  * Payment receipt claims now reject empty payment option identifiers.
  * Validation behavior is consistent across supported schema formats, helping prevent incomplete payment data from being accepted.

* **Tests**
  * Added coverage confirming empty required fields are rejected across payment requests, options, and receipt claims.

### DIFF
--- a/.changeset/ack-pay-reject-empty-payment-fields.md
+++ b/.changeset/ack-pay-reject-empty-payment-fields.md
@@ -1,0 +1,7 @@
+---
+"@agentcommercekit/ack-pay": patch
+---
+
+Reject empty payment request and payment option identifiers.
+
+`paymentOptionSchema` accepted empty strings for `id`, `currency`, and `recipient`, and `paymentRequestSchema` / `paymentReceiptClaimSchema` accepted empty request and option ids. Require non-empty strings for these fields in both the Valibot and Zod schemas so HTTP 402 bodies and verified payment request tokens cannot carry blank identifiers.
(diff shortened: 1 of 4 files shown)


venables:
Thank you for the contribution. This duplicates #183, which was opened first and is in review. Contributing to that PR is more helpful than a parallel version. Closing as a duplicate.

PR #220, opened 2026-09-18
fix(ack-pay): reject empty payment option network

### LINKED ISSUE #219: bug(ack-pay): paymentOptionSchema accepts an empty network string
## Summary

`paymentOptionSchema` treats `network` as optional, but when the field is present both schemas still accept an empty string:

- valibot: `network: v.optional(v.string())`
- zod: `network: z.string().optional()`

So `{ ..., network: "" }` parses successfully. An empty network is not a meaningful CAIP / chain hint; it is the same class of empty-field hole that #183 covers for `id` / `currency` / `recipient`, just on the optional field.

## Affected files

- `packages/ack-pay/src/schemas/valibot.ts`
- `packages/ack-pay/src/schemas/zod.ts`

## Expected

If `network` is present, require a non-empty string (same `nonEmptyString` / `min(1)` shape as the other identifier fields). Omitting the field remains valid.

## AI usage

Filed with Cursor assistance while scanning for schema gaps not covered by open PRs. I reviewed the schema lines above and take responsibility for the report.

kutluhaneth46:
## Summary
- Fixes #219: `paymentOptionSchema` accepted `network: \"\"` even though an empty network is not a meaningful chain hint.
- Valibot uses `v.optional(nonEmptyString)`; zod uses `z.string().min(1).optional()`. Omitting `network` remains valid.
- Parity tests cover omitted, non-empty, and empty `network` on both schemas.

## AI usage
Cursor assisted locating the gap, writing the schema change, parity tests, and changeset. I reviewed and understand the change: when `network` is present it must be non-empty; absence is still allowed.

## Test plan
- [x] `pnpm --filter @agentcommercekit/ack-pay exec vitest run src/schemas/payment-option.test.ts` — 19/19 pass

## Summary by CodeRabbit

- **Bug Fixes**
  - Payment options now reject an empty `network` value when the field is provided.
  - A missing `network` value remains valid.
  - Non-empty network names continue to be accepted consistently across supported validation methods.

- **Tests**
  - Added coverage confirming validation behavior for omitted, empty, and non-empty `network` values.


venables:
Thank you for the contribution. Empty-string handling for payment option fields is already being addressed in #183, which was opened first. An empty network string belongs in that same change rather than a separate PR, so I am closing this one and folding the case into #183.

kutluhaneth46:
Understood — thanks for pointing at #183. Happy to help there if anything is still open.

PR #221, opened 2026-09-18
fix(vc): fail closed when expirationDate is unparseable

### LINKED ISSUE #148: bug: `isExpired` returns `false` for unparseable `expirationDate` (fail open)

kutluhaneth46:
## Summary

`isExpired` returned `false` for a present but unparseable `expirationDate`, so a malformed expiry silently passed verification. Status-list checking already fails closed for the same case.

## Changes

- Treat unparseable (and empty) `expirationDate` values as expired
- Expand unit coverage for the fail-closed path
- Refresh the status-list comment and README note
- Add a patch changeset for `@agentcommercekit/vc`

## AI disclosure

Assisted with Cursor. I reviewed and understand the fail-closed change and its interaction with `verifyParsedCredential` and the status-list expiry path.

Closes #148

## Summary by CodeRabbit

* **Bug Fixes**
  * Verifiable Credentials with invalid, empty, non-ISO, or calendar-overflow expiration dates are now treated as expired, supporting fail-closed verification.
  * Expiration checks now correctly distinguish invalid calendar dates from valid timestamps with timezone offsets.
  * Dates that appear valid but are silently normalized by date parsing are treated as expired.

* **Documentation**
  * Clarified in the API reference that unreadable or malformed expiration dates are treated as expired.


kutluhaneth46:
Addressed the overflow-calendar edge case from review: ISO-shaped dates like `2099-02-30T00:00:00.000Z` that `Date` would normalize are now treated as expired, with a regression test. `is-expired` tests 8/8 green.

kutluhaneth46:
Follow-up for the calendar-bounds review note: overflow detection now checks year/month/day against calendar length instead of UTC fields, so valid offsets like `2099-01-01T00:00:00+14:00` stay non-expired. Regression coverage added; `is-expired` tests 9/9 green.

venables:
Thank you for the fix. This change is already open in #154, which was opened first and has been reworked after review. Please contribute there instead of a parallel version. Closing as a duplicate of #154.

kutluhaneth46:
Thanks for the pointer to #154 — will follow up there instead.

